PR #760, opened 2013-11-22
Improve processing of machine check exceptions. Resolves rhbz812537.

dvlasenk:
MCE's are detected by abrt as kernel oopses,
but they are hardware errors, not kernel bugs.

This change makes it more clear for the user what they are,
shows more information if it is available, otherwise
explains to user what needs to be done to get it -
user needs to install mcelog tool.

If non-fatal MCE is seen, abrt will detect it as an oops
and alert user in a usual manner. When user opens this
abrt problem for reporting, he will see that "comment"
field is pre-filled with some text.
What it says depends on whether mcelog tool is installed.

If mcelog is installed, the text is:
"""
The kernel log indicates that hardware errors were detected.
/var/log/mcelog file may have more information.
# The last 20 lines of /var/log/mcelog are:

<up to 20 lines of log>
"""

otherwise it is:

"""
The kernel log indicates that hardware errors were detected.
The data was saved by kernel for processing by the mcelog tool.
However, /var/log/mcelog file does not exist.
Most likely reason is that mcelog is not installed or not configured
to be started during boot.
Without this tool running, the binary data saved by kernel
is of limited usefulness.
(You can save this data anyway by running 'cat </dev/mcelog >FILE').
The recommended course of action is to install mcelog.
If another hardware error would occur, a user-readable description
of it will be saved in /var/log/mcelog.
"""

I hope the message is clear enough to let user understand that
it is a hardware error, and instead of filing a bug and/or
talking to our support, user will start looking at his hardware.

If fatal MCE is encountered, kernel always panics,
(abrt has no chance of catching the oops),
kdump kicks in, and then after reboot abrt says that new vmcore
is found. When user generates backtrace, he will see oops text
which starts with
"Machine Check Exception: BANK nnn ...".

(Yes, it's weird that kernel shows human-readable error messages
on fatal MCEs but doesn't do that for non-fatal ones.
This makes fetching MCE info significantly different...
I wish kernel would show human-readable MCEs in both cases,
we wouldn't need mcelog then... oh well.)

This change was run-tested on RHEL6 machine.

Signed-off-by: Denys Vlasenko dvlasenk@redhat.com

### DIFF
--- a/src/lib/kernel.c
+++ b/src/lib/kernel.c
@@ -17,6 +17,7 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
+#define _GNU_SOURCE 1 /* for strcasestr */
 #include "libabrt.h"
 
 /*
@@ -27,6 +28,11 @@ struct line_info {
     char *ptr;
     char level;
 };
+
+/* Used to be 100, but some MCE oopses are short:
+ * "CPU 0: Machine Check Exception: 0000000000000007"
+ */
+#define SANE_MIN_OOPS_LEN 30
 
 static void record_oops(GList **oops_list, struct line_info* lines_info, int oopsstart, int oopsend)
 {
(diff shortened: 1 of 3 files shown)


trapas:

trapas:

juliusmilan:
Closing, because we are no longer adding new features into rhel6 and without customer requirement till rhel6 EOL, this patch probably won't be backported.

(Present in upstream)

PR #903, opened 2015-01-20
Use GSettings to check whether to send uReports

jfilak:


hadess:
That seems counter-productive to me. That would mean that sysadmins need to change 2 settings if they want to force having uReports sent, instead of one, depending on the desktop to be used.

Catanzaro:
It might make for sense for gnome-control-center to just parse the existing configuration file, if you don't want to switch fully to gsettings.

Catanzaro:
I mean, it'd be NICE to have this as a gsetting, since that's where all our other settings are, but having the same setting in two places seems worse to me than the status quo.

jfilak:
Since I am not able to determine whether the gsetting has been set, I had to use this approach. I would rather use the gsetting only, but I want to display the dialogue in non-GNOME desktops and update the gsetting from the applet as the other desktops don't have their privacy panels yet.

hadess:
Why do you need to register whether the GSettings had been set? We want to remove the nag dialogue in GNOME. I'm this close -><- to ending the argument by forking the applet. There's no reason to be spending this much time to engineering 1000 lines of code.

jfilak:
@hadess OK, the dialogue will not be removed in GNOME.

PR #906, opened 2015-01-21
Console notification timeout

### LINKED ISSUE #898: login stucked because of .lock file
Login didn't work anymore because of this file:
/var/spool/abrt/ccpp-2015-01-14-14:19:56-26366/.lock

We have abrt-console-notification installed, which tries to run "abrt-cli status" on every login. Because of the above file, the command was never finishing. This is the output from strace:

nanosleep({0, 500000000}, NULL)         = 0
symlink("4729", "/var/spool/abrt/ccpp-2015-01-14-14:19:56-26366/.lock") = -1 EEXIST (File exists)
readlink("/var/spool/abrt/ccpp-2015-01-14-14:19:56-26366/.lock", "21864", 15) = 5
access("/proc/21864", F_OK)             = 0
nanosleep({0, 500000000}, NULL)         = 0
symlink("4729", "/var/spool/abrt/ccpp-2015-01-14-14:19:56-26366/.lock") = -1 EEXIST (File exists)
readlink("/var/spool/abrt/ccpp-2015-01-14-14:19:56-26366/.lock", "21864", 15) = 5
access("/proc/21864", F_OK)             = 0

Maybe the command should timeout in a few seconds and an error message displayed in this case?

jfilak:
Closes #898

### DIFF
--- a/src/cli/abrt-console-notification.sh
+++ b/src/cli/abrt-console-notification.sh
@@ -39,4 +39,4 @@ if [ -f "$TMPPATH" ]; then
     mv -f "$TMPPATH" "$SINCEFILE" >"$ABRT_DEBUG_LOG" 2>&1
 fi
 
-abrt-cli status --since="$SINCE" 2>"$ABRT_DEBUG_LOG"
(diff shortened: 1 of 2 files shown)


mmilata:
Works fine, merged.

Issue #907, opened 2015-01-26
Add "binary path" to Crash() signal


hadess:
Which would do the equivalent of:
```
 struct dump_dir *dd = dd_opendir($DIR_PATH, DD_OPEN_READONLY);
 char *value = dd_load_text_ext(dd, FILENAME_EXECUTABLE, DD_FAIL_QUIETLY_ENOENT | DD_LOAD_TEXT_RETURN_NULL_ON_FAILURE);
```

This would be useful to implement the notification mockups: "We're sorry, it looks like Foobar crashed".
It would also, when binaries inside a package that's not the application crashes avoid offering to restart the application.

jfilak:
Hmm, don't you need 'full command line' to be able to find a proper destkop file?

hadess:
We shouldn't need the full command-line, no. Why would we?

jfilak:
A single binary can be used in many desktop files. See the discussion here : https://github.com/abrt/gnome-abrt/issues/76

hadess:
Right, except that "/usr/bin/nautilus --g-application-service" won't correspond to any desktop files, and "/usr/bin/gnome-control-center wacom" should correspond to "Settings", not the displays .desktop entry.
We could deal with further problems on a case-by-case basis.

jfilak:
So why not to add both and be ready for the further problems ;)

hadess:
Sure. Or you could make crash send an a{sv} (an array of string/values) so that we can extend the API without breaking it and send both.

jfilak:
I'm sorry, this isn't going to happen. Please use the ``dd_opendir()`` approach (see  #908 for more details).

PR #908, opened 2015-01-26
a-a-notify: send cmdline and executable in Crash

jfilak:
Related to #907


jfilak:
It turned out we need more fields than we thought and adding all the required fields to the Crash signal seems to be superfluous, because we can read them from the related problem directory.

PR #931, opened 2015-03-04
Python 3 compatibility

mmilata:
This patch just applies what 2to3 found so that the scripts can run both
under python 2 and python 3. More work is needed to fully switch to
python 3.

Related to rhbz#1192080.

Signed-off-by: Martin Milata <mmilata@redhat.com>


jfilak:
Thank you!

PR #935, opened 2015-03-19
abrt-applet: get new problems over DBus

jfilak:
Without this pull request, abrt-applet cannot single out problems to notify at start-up because the current approach is to iterate through `ls DUMP_LOCATION` but the dump location is not accessible for regular users.


jfilak:
Version 2:
- removed ```GList **```
- no initialization of ```GAppinfo *app``` to NULL

jfilak:
I squashed the 1st and the 3rd commit and pushed the resulting two patches to the master branch.

PR #945, opened 2015-03-30
fix kernel oops parsing for kernel >= 4

jfilak:
ABRT cannot process any kernel oops produced by kernel > 3 without this pull request.

The journal-oops-processing and oops-processing tests covers this PR.


jfilak:
Thank you! I've  added escaping for the dots and pushed the updated version to master.

PR #952, opened 2015-04-16
Ccpp vulnerabilities

jfilak:
Requires: abrt/libreport#346
Requires: abrt/satyr#225

### DIFF
--- a/abrt.spec.in
+++ b/abrt.spec.in
@@ -487,11 +487,7 @@ to the shell
 %build
 autoconf
 
-%if 0%{?rhel:%{rhel} < 7}%{?fedora:%{fedora} < 18}
-    %define var_base_dir spool
-%else
-    %define var_base_dir tmp
-%endif
+%define var_base_dir spool
 
 CFLAGS="%{optflags} -Werror" %configure \
 %if 0%{?suse_version}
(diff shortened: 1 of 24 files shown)


sorki:
Care to explain how does 217e822 (ccpp: create dump directory without parents) makes the code safer?

sorki:
I think it would be better if 96a1bb0 is split into two patches - one fixing simlink races and the oher one disabling java log saving.

sorki:
@trapas test this please

mbrysa:
#985 abrt-dbus should be restarted after abrt package update, otherwise it keeps loading dumpdirs from `/var/tmp/abrt` until killed.

mbrysa:
abrt-cli must do chowning before reporting. Otherwise regular users cannot report koopes via cli.

mbrysa:
After trying to report a koops using gnome-abrt as regular user, the DD is properly chowned to user:abrt, but the `.libreport/owner` still contains uid of nobody.

mbrysa:
gnome-abrt must use d-bus to access problem details. Related https://github.com/abrt/gnome-abrt/issues/64 .

mbrysa:
All the issues I pointed out are fixed in the latest rebase.

mbrysa:
Thank you!

PR #953, opened 2015-04-17
RHEL6: ccpp: do not log to syslog while dumping syslog

jfilak:
If the crashing process has the socket /dev/log, then turn off logging to
syslog.

Instead of turnning the logging off completely abrt-hook-ccpp switches to
logging to STDIO for sake of debugging.

Resolves: #1199261


jfilak:
rhel6 commit 0b51a3d7f068c57c1d87c787253b819a213b95bf

PR #955, opened 2015-04-20
upload: validate and sanitize uploaded dump directories

jfilak:
It was discovered that, when moving problem reports from
/var/spool/abrt-upload to /var/spool/abrt or /var/tmp/abrt,
abrt-handle-upload does not verify that the new problem directory
has appropriate permissions and does not contain symbolic links.  A
crafted problem report exposes other parts of abrt to attack, and
the abrt-handle-upload script allows to overwrite arbitrary files.

Acknowledgement:

This issue was discovered by Florian Weimer of Red Hat Product Security.

Related: #1212953

### DIFF
--- a/src/daemon/abrt-handle-upload.in
+++ b/src/daemon/abrt-handle-upload.in
@@ -10,6 +10,7 @@ import getopt
 import tempfile
 import shutil
 import datetime
+import grp
 
 from reportclient import set_verbosity, error_msg_and_die, error_msg, log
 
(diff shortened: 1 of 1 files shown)


jfilak:
Thank you @fweimer! I've amended the commit. Here is the old commit with comments: https://github.com/abrt/abrt/commit/f65c2f4a4f5dd031c2ec82bd458a0e6d018f8fd4

PR #956, opened 2015-04-22
Port the applet to D-Bus methods

jfilak:
This patch is a part of our efforts to make abrt-applet independent on
the backend.

This patch converts all data manipulation functions to D-Bus calls, so
the notifications are made of data obtained through D-Bus.

The reporting still relies on file system access, though.

### DIFF
--- a/src/applet/applet.c
+++ b/src/applet/applet.c
@@ -120,6 +120,7 @@ typedef struct problem_info {
     bool reported;
     bool was_announced;
     bool is_writable;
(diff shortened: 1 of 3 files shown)


jfilak:
I've fixed the missing element flaw found in commit dc02b39735acf2707c577e0b4bad8e5fb190fad5 and amended this pull request.

mhabrnal:
Looks fine now, Thanks!

Issue #957, opened 2015-04-24
abrt-auto-reporting does not work on Fedora and CentOS


mtoman:
`abrt-auto-reporting` tries to load `rhtsupport.conf` even on Fedora and CentOS and dies when the file does not exist. This makes `abrt-auto-reporting` only usable on RHEL.

PR #958, opened 2015-04-24
require rhtsupport.conf file only on RHEL

mhabrnal:
abrt-auto-reporting required the rhtsupport.conf on Fedora and CentOS but the conf
file do not exists in those systems.

Resolves abrt/abrt#957


jfilak:
`abrt-auto-reporting --help` still includes options used for the authentication.

mhabrnal:
I pushed a new version to this pull request, I've added man pages too.

PR #962, opened 2015-05-05
don't run whole test if reporter-mantisbt isn't installed

mhabrnal:


sorki:
You can use shorter form:

rpm -q package || rlDie "msg"

sorki:
Thanks!

PR #967, opened 2015-05-14
abrt-action-list-dsos: dirty packages don't have a RPMTAG_VENDOR item

mhabrnal:
In case there are installed some dirty packages the abrt-action-list-dsos failed.

### DIFF
--- a/src/plugins/abrt-action-list-dsos
+++ b/src/plugins/abrt-action-list-dsos
@@ -82,10 +82,15 @@ if __name__ == "__main__":
                         if outname:
                             outfile = xopen(outname, "w")
                             outname = None
+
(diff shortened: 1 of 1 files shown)


jfilak:
"dirty packages" is not a general term. ABRT's gen-version scripts add "-dirty" suffix to Release if your local copy repository has not-yet commited changes. While the patch looks almost good (the construction you have used is confusing), the description is totally wrong :) because there are no "dirty packages".

h[rpm.RPMTAG_VENDOR]  was always None for packages build using "make rpm" and rpmlib is happy with None in vedora param but commit 257bee0b added .decode('utf-8') which is the root cause.

mhabrnal:
I've addressed your comments and I rebased the pull request.

PR #968, opened 2015-05-17
GSettings integration

### LINKED ISSUE #966: ABRT settings should be in sync with Privacy settings in Control Center
Even if I have automatic reporting enabled in ABRT, reports are not reported because it's overrode by settings in Privacy in Control Center. IMHO those two settings should be in sync, otherwise it's very confusing for users.

jfilak:
Closes #966

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -3,6 +3,7 @@
 # Please keep this file sorted alphabetically.
 src/applet/abrt-applet.desktop.in
 src/applet/applet.c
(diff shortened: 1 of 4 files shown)


mhabrnal:
It's not possible to make rpm. Adding 'src/configuration-gui/abrt-config-widget.c' to POTFILES.in should fix the issue.

jfilak:
D'oh, fixed.

mhabrnal:
Works for me, thanks!

PR #970, opened 2015-05-18
applet: fix refcount problem info mechanism

mhabrnal:
Fix a double free problem.

Related to rhbz#1211644


jfilak:
Sounds like a good idea but you increment the reference counter in a wrong place. If you take a closer look you will see that you does "++ref; something not modifying ref; --ref" and the result is nothing. I don't think there is not enough ref calls but unref is called more than necessary.

mhabrnal:
Good point, you are right. I've rebased the pull request.

jfilak:
Thank you!

PR #974, opened 2015-05-22
testsuite: add authenticated autoreporting by certificate test

mhabrnal:
The first part tests abrt-auto-reporting sanity which not contains an
authenticated options. This part is the same for fedora and rhel and runs on both.
The second one tests options for authenticated reporting and the test runs only
on rhel.

Related to rhbz#1223822


trapas:

sorki:
@trapas test this please

PR #975, opened 2015-05-27
Abrt python fixes

sorki:
Few improvements and bugfixes. Related to new cli.

One unrelated cli commit removing dead code included.

### DIFF
--- a/src/cli/abrt-cli.c
+++ b/src/cli/abrt-cli.c
@@ -75,25 +75,6 @@ static unsigned handle_internal_options(int argc, const char **argv, const char 
         {
             return skip + argc;
         }
-#if 0
-        if (prefixcmp(cmd, "--base-dir=") == 0)
-            D_list = g_list_append(D_list, xstrdup(cmd + strlen("--base-dir=")));
-        else if (prefixcmp(cmd, "--list-events") == 0)
-        {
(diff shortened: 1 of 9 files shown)


jfilak:
Looks good! ACK from me. @mbrysa, can you please review the Python code?

sorki:
- moved id and short_id from utils to Problem class as properties
- python3 compat - hashlib instead of sha
- added test for new properties

jfilak:
I just want to make sure you intentionally omitted tests for `non_reportable` and `non_reportable_reason`.  Or those things are not properties?

sorki:
Will add these. They are trivial but nice to have anyway..

sorki:
Fixed both issues. Added setters for non_reportable properties

sorki:
Removed leftover print

mbrysa:
Why is the property for `not-reportable` called `non_reportable`?

sorki:
not-reportable is not syntactically acceptable in python

sorki:
also it should be renamed to not_reportable to comply with other elements like reported_to

sorki:
sed s/non_reportable/not_reportable/g

mbrysa:
Thank you!

PR #981, opened 2015-05-29
cli-ng

sorki:
abrt-cli rewrite adding additional functionality like abrt gdb, abrt debuginfo-install, tab-completion and custom formatting. Feature parity with old cli remains almost unchanged.

Should replace old cli in the future.

Depends on #975. (cli-ng branch contains commits from both pull requests).

Missing stuff:
- abrt-cli process (do we want this one? it does interactive processing of all problems like git add -i)
- man page
- i18n
- not all variables (paths) in config.py.in replaced during build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 src/applet/abrt-applet
 src/configuration-gui/system-config-abrt
 src/daemon/abrt-action-save-package-data
+src/daemon/abrt-action-save-container-data
 src/daemon/abrt-auto-reporting
 src/daemon/abrt-server
 src/daemon/abrtd
@@ -37,12 +38,14 @@ src/plugins/abrt-action-perform-ccpp-analysis
 src/plugins/abrt-action-trim-files
 src/plugins/abrt-watch-log
 src/plugins/abrt-dump-oops
-src/plugins/abrt-journal-dump-oops
+src/plugins/abrt-dump-journal-oops
+src/plugins/abrt-dump-journal-core
 src/plugins/abrt-dump-xorg
 src/plugins/abrt-retrace-client
 src/plugins/abrt-action-install-debuginfo
 src/plugins/abrt-action-generate-core-backtrace
 src/plugins/abrt-bodhi
+src/plugins/libabrt-journal.a
 src/python-problem/problem/config.py
 src/python-problem/tests/*.log
 src/python-problem/tests/*.trs
@@ -165,3 +168,9 @@ tests/package.m4
 # python API docs builddir
 src/python-problem/doc/_build/
 build/*
+
+# python tests and generated files
+*.trs
+.coverage
+toolswtf
+src/cli-ng/abrtcli/config.py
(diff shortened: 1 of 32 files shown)


sorki:
I'll finish all the missing stuff but I wanted to have this out now for testing and discussion.

sorki:
@trapas test this please

sorki:
Forgot to mention that this takes prominent 'abrt' binary name instead of 'abrt-cli-ng' or a'abrt-cli' to save people some typing.

sorki:
Fixed tests. They won't pass in this PR until #975 is merged but cli-ng branch now builds fine.

sorki:
Copr build: https://copr.fedoraproject.org/coprs/rmarko/abrt/

mhabrnal:
Aliases for action parameters don't work for me ('abrt ls' etc.)

sorki:
Yup, aliases are only supported in new version of python argh available for python 3. They'll start working as soon as we switch to py3 or with new version of python argh.

cli-ng can't run py3 as of now due to missing python3-argcomplete (https://bugzilla.redhat.com/show_bug.cgi?id=1225934).

```
This decorator only works with a recent version of argparse 
(see Python issue 9324 and Python rev 4c0426). 
Such version ships with Python 3.2+ and may be available
in other environments as a separate package. 
Argh does not issue warnings and simply ignores aliases if they are not supported.
```

sorki:
sed s/non_reportable/not_reportable/g

sorki:
Do not merge yet. Still few minor fixes pending + there will possibly be additional stuff from todays announcements.

sorki:
Ready for review. Added man page, authentication support and i18n support.

jfilak:
Please call the ChownProblemDir DBus method to allow non-privileged users to read files in the problem directories:
https://github.com/abrt/gnome-abrt/blob/master/src/gnome_abrt/dbus_problems.py#L145
https://jfilak.fedorapeople.org/ProblemsAPI/re01.html#org.freedesktop.Problems.ChownProblemDir

We will get rid of the need to chown the problem directories in the Problems API v2.

sorki:
Ping. Anyone?

sorki:
@trapas test this please

sorki:
Added integration test

sorki:
@trapas test this please

sorki:
Updated to use ask_yes_no from libreport and fetch username from problem dir instead of lookup on current system. I won't add any missing functionality (including python bindings) to libreport.

jfilak:
The following comment has been removed by "git rebase":
https://github.com/abrt/abrt/commit/19950d2aab422c549213d74a362b172e93d09fee#commitcomment-13561830

And this is the commit with my review comments:
https://github.com/abrt/abrt/commit/ee624b8919a02b65142b42eafc25863850d5b037

sorki:
Build IDs fixed in libreport PR#384.

Fixed tests, also running on PY3.

sorki:
added py3 build deps

jfilak:
Thank you! Good job!

PR #991, opened 2015-06-08
abrt: Fixup component of select kernel backtraces

mbrysa:
The kernel is a big project and certain parts of it
may need to be tracked under different components.
Fixup results related to those parts and assign a
different component.

Signed-off-by: Laura Abbott <labbott@fedoraproject.org>


mbrysa:
This must be tested with FAF to make sure it doesn't break uReporting.

jfilak:
I fixed Marek's findings and added a test case.

mbrysa:
Thank you!

PR #1001, opened 2015-06-23
C/C++: Thread ID

jfilak:
Save TID and use it do get the crash thread. Requires: abrt/libreport#369


mhabrnal:
Looks good for me, thanks!

PR #1005, opened 2015-07-01
cli: enable polkit authentication on command line

jfilak:
This patch will allow users to work with all problems without the need
to run abrt-cli under root account.

The polkit aget will run in a separate thread and will interact with a
user via STDOUT and STDIN, so we should not introduce new threads using
STDIN or STDOUT and all D-Bus calls should be synchronous.

http://www.freedesktop.org/software/polkit/docs/latest/ref-authentication-agent-api.html


mhabrnal:
Looks good

PR #1009, opened 2015-07-08
Enable authetictaion in 'abrt-cli process"

jfilak:


mhabrnal:
works for me, merged.

PR #1010, opened 2015-07-13
a-dump-oops: allow update the problem, if more then one oops found

mhabrnal:
In case that found more than one oops process the first one.
Without this patch the script exits with error in this case because expects
only one oops.

Related to rhbz#1170534


jfilak:
The code changes look good to me, but would it be possible to add a test case?

mhabrnal:
Add the test case.

PR #1011, opened 2015-07-14
Switch to Python 3

jfilak:


mhabrnal:
Package python3-systemd doesn't exist on Fedora 22, can you do the spec file adaptive?

jfilak:
Sure, I was thinking about that but it was so easy to push this version, but I have to admit that it was a bad idea. I will not be so lazy next time.

jfilak:
Hmm, I forgot that there is no python3-dmidecode in Fedora 22, so I cannot make the Python3 requires conditional. Sorry, you have to test it on Rawhide.

mhabrnal:
I tested it on Rawhide and It looks good to me.

PR #1012, opened 2015-07-15
abrtd: de-prioritize post-create event scripts

jfilak:
The crash processing should not make the computer unusable. It sometimes
happens that the captured data causes abrt scripts to take an inadequate
amount of resources and the computer becomes less responsive.

This patch increases the nice value of post-create processes by 10 (I took
10 because it is the default value of command 'nice'), so those
processes will be scheduled after the more valuable processes.

Related: rhbz#1236422

### DIFF
--- a/doc/abrtd.txt
+++ b/doc/abrtd.txt
@@ -36,6 +36,13 @@ OPTIONS
 -p::
    Add program names to log.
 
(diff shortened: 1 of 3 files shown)


mhabrnal:
looks good to me, thanks

PR #1018, opened 2015-07-28
abrt-auto-reporting: fix related to conditional compilation

mhabrnal:
We discovered that conditional compilation in abrt-auto-reporting does not
work. We forgot add -DAUTHENTICATED_AUTOREPORTING=1 flag if
AUTHENTICATED_AUTOREPORTING is enabled.

Related to rhbz#1191572


jfilak:
Thanks!

PR #1027, opened 2015-08-12
testsuite: abrt-auto-reporting-sanity - exclude one phase on RHEL

optak:
phase "missing rhtsupport.conf" is not relevant for RHEL


trapas:

jfilak:
Thank you!

PR #1031, opened 2015-08-24
testsuite: useless var_log_messages files are not created

mhabrnal:
In the case, there are useless log messages the var_log_messages file is not
created in dump dir.

Related to rhbz#1255411


mhabrnal:
I've addressed your comment and rebased the pull request.

PR #1032, opened 2015-08-24
testsuite: d-e-handling set 'PrivateReports' to 'no' in abrt.conf

mhabrnal:
Dump directories are owned by root now (introdiced in commit
https://github.com/abrt/abrt/commit/8939398b82006ba1fec4ed491339fc075f43fc7c).
Due to this changes the test began to fail. We want to regular users to
own the problem data colleted by abrt in the dbus-element-handling test.
To ensure this, we must set the 'PrivateReports' option to 'no' in abrt.conf.
The 'PrivateReports' option is only on RHEL.


mhabrnal:
Updated

PR #1045, opened 2015-09-15
Rhel6 tests branched

sorki:


jfilak:
Cool! Good job! I just don't understand why we are dropping "console-notification: drop time out test" instead of fixing it. I remember yesterday's IRC discussion but I told you that we should fix the test (either by reverting the commit that made the test "useless" (if I use your words) or rewriting it from scratch). I do not want to drop the test because rhel-6 ships with abrt-console-notifications and I want to be sure that abrt does not break the log in.

PR #1046, opened 2015-09-15
Rhel7 tests branched

sorki:


sorki:
added missing tests

jfilak:
Thank you very much!

PR #1052, opened 2015-09-24
runtests: stick to new BZ password rules

martinky82:
With the recent bugzilla update, new more strict password rules are enforced. The testing account pw has been changed, thus the tests need to use it.


trapas:

sorki:
Thanks!

jfilak:
Don't we need to cherry-pick the commit to `rhel7` and `rhel6` branches too?

martinky82:
Yes, we do

PR #1053, opened 2015-09-24
testsuite: use rpm to remove packages

sorki:
dnf dependens on libreport-filesystem
and uninstall itself when trying to remove libreport-filesystem.

Signed-off-by: Richard Marko <rmarko@fedoraproject.org>


jfilak:
Thank you!

PR #1054, opened 2015-09-29
testsuite: ureport-auth: force cp/mv when restoring configuration

martinky82:
When running the test by hand, the configuration doesn't get restored properly. This simple patch makes restoring more robust.


trapas:

jfilak:
Thank you!

PR #1057, opened 2015-10-08
abrt-retrace-client: use atoll for _size conversion

sorki:
Otherwise we get

The size of your crash is 28.1 MiB, but the retrace server only accepts
crashes smaller or equal to 16.0 EiB.

(gdb) print settings->max_unpacked_size
$4 = -50331648


jfilak:
Good catch! Thank you!

PR #1065, opened 2015-10-13
logging: spit messages on start/end of processing

sorki:
Otherwise ccpp hook is too quiet and it takes
a while until post-create logs anything.

### DIFF
--- a/src/hooks/abrt-hook-ccpp.c
+++ b/src/hooks/abrt-hook-ccpp.c
@@ -623,6 +623,11 @@ int main(int argc, char** argv)
                         (long)pid, executable);
     }
 
+    const char *signame = NULL;
+    const int fatal_signal = signal_is_fatal(signal_no, &signame);
+
+    log_notice("Processing '%s' got signal SIG%s", executable, signame);
+
     user_pwd = get_cwd(pid); /* may be NULL on error */
     log_notice("user_pwd:'%s'", user_pwd);
 
@@ -665,9 +670,10 @@ int main(int argc, char** argv)
         return create_user_core(user_core_fd, pid, ulimit_c);
     }
 
-    const char *signame = NULL;
-    if (!signal_is_fatal(signal_no, &signame))
-        return create_user_core(user_core_fd, pid, ulimit_c); // not a signal we care about
+    if (!fatal_signal) { // not a signal we care about
+        log_debug("Non-fatal signal, ignoring");
+        return create_user_core(user_core_fd, pid, ulimit_c);
+    }
 
     const int abrtd_running = daemon_is_ok();
     if (!setting_StandaloneHook && !abrtd_running)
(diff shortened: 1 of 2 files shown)


sorki:
Added singal

abrt-hook-ccpp[2735]: Processing '/usr/bin/will_segfault' got signal SIGSEGV

'Done processing' is useful by itself for timing how long the processing chain takes.

I'm working on adding useful logging to run-event (which event failed file/line number).

jfilak:
> 'Done processing' is useful by itself for timing how long the processing chain takes.

OK. But it is useful for us when we are debugging abrt and not for regular users, right?

sorki:
Change level of 'Done proc' to debug

src/hooks/abrt-hook-ccpp.c
```diff
@@ -665,9 +670,10 @@ int main(int argc, char** argv)
         return create_user_core(user_core_fd, pid, ulimit_c);
     }
 
-    const char *signame = NULL;
-    if (!signal_is_fatal(signal_no, &signame))
-        return create_user_core(user_core_fd, pid, ulimit_c); // not a signal we care about
+    if (!fatal_signal) { // not a signal we care about
```
mhabrnal: Use our standard code style.
if ()
{
...
}

src/hooks/abrt-hook-ccpp.c
```diff
@@ -623,6 +623,11 @@ int main(int argc, char** argv)
                         (long)pid, executable);
     }
 
+    const char *signame = NULL;
+    const int fatal_signal = signal_is_fatal(signal_no, &signame);
+
+    log_notice("Processing '%s' got signal SIG%s", executable, signame);
```
jfilak: If `signal_no` is not a fatal one, `signame` is `NULL` (`signal_is_fatal()` function does not translate all signal numbers to human readable strings), so you will see message like this in the logs:
`Processing '/usr/bin/foo' got signal SIG(null)`

How about to move `reason` from line 848 to above this line and use `reason`  for the logging purpose too?

`reason` doesn't contain full path but I think we don't need to included this information in the message. This is how dmesg line looks like when a process gets SEGV:
`[ 5059.103443] will_segfault[7418]: segfault at 0 ip 000055851b95fa00 sp 00007ffd2c631120 error 4 in will_segfault[55851b95f000+1000]`
The message is rather strange, but does not contain the full path. Also If I run a poor quality code written by myself, I will see very long log messages, because those binaries are stored in a deep sub-directory structure in my home directory.

Even `systemd-coredump` includes only the base name:
`Oct 14 07:43:14 photon systemd-coredump[7588]: Process 7587 (will_segfault) of user 18601 dumped core.`

Speaking of `systemd-coredump` , no one is complaining about spamming system logs, so be progressive and use `log` instead of `log_notice` (perhaps you can log it as ERROR, to see a nice, red colored message in journald). I wanted to do that but I didn't have time to come up with a good plan for better logging (I am burden with D-Bus API v2, CVEs and bureaucracy, you know), so I have been keeping status quo since I silenced `abrt-hook-ccpp` on William Jon McCann's request.

jfilak:
Thank you for the pull request. Commit https://github.com/abrt/abrt/commit/a72d05b10deb6303440f8a63c729d19d214ca964 implements your idea.

PR #1067, opened 2015-10-13
remove random code example from abrt-server

sorki:


jfilak:
OK, push commits like this one without opening PRs, please.

PR #1072, opened 2015-10-16
testsuite: upload-handling: fix irrelevant AVCs

martinky82:
using cp instead of mv ensures the tarball to have proper SELinux context


trapas:

jfilak:
Thanks!

PR #1073, opened 2015-10-16
testsuite: dumpdir_compltedness: verify sosreport contents

martinky82:
Verify contents of the sosreport tarball (normalized)


trapas:

martinky82:
It seems this feature is quite complex and is broken as is in this patch. Dropping for now.

PR #1076, opened 2015-10-26
a-d-journal-core: set root owner for created dump directory

mhabrnal:
Without this commit abrt-server fails during check the right owner and group of
dump dirs.

Related to rhbz#1269827


jfilak:
Good job! Please add a test case for abrt-dump-jornal-core (that was my job but I probably forgot about it, sorry).

mhabrnal:
I've added `journalctl` to the man page and added test case for abrt-dump-journal-core.

jfilak:
Thank you!

PR #1086, opened 2015-11-11
C/C++: processes with locked memory

### LINKED ISSUE #796: ABRT should consider processes with locked memory security sensitive
I was quite surprised when I noticed keepassX reports from ABRT. It's a password manager and thus very likely to contain secrets in tracebacks.

A good way to tell a secret manager from other processes is by looking whether it has any memory locked (VmLck != 0 kB in /proc/<pid>/status).

Such process should either be removed from ABRT consideration or treated as potentially containing very sensitive information.

jfilak:
Closes #796


mhabrnal:
Looks good to me. Thank you!

PR #1090, opened 2015-11-24
runtests: ccpp-plugin-debug: normalize paths for greping

martinky82:
get rid of double slashes


trapas:

PR #1098, opened 2015-12-14
ccpp: use executable name from pid

mhabrnal:
Using executable name from core_pattern's parameter is not safe because in some
cases the core_pattern can contain bogus "executable name" parameter.

See
https://github.com/abrt/abrt/blob/master/src/hooks/abrt-install-ccpp-hook.in#L16

### DIFF
--- a/src/hooks/abrt-hook-ccpp.c
+++ b/src/hooks/abrt-hook-ccpp.c
@@ -519,13 +519,16 @@ static void error_msg_process_crash(const char *pid_str, const char *process_str
     char *message_full = xvasprintf(message, p);
     va_end(p);
 
-    if (signame)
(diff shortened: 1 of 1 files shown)


src/hooks/abrt-hook-ccpp.c
```diff
@@ -646,6 +646,9 @@ int main(int argc, char** argv)
     pid_t pid = xatoi_positive(argv[8]);
 
     char *executable = get_executable(pid);
+    const char *last_slash = strrchr(executable, '/');
```
hberan: What happens if get_executable returns NULL?

mhabrnal:
Good point, I've addressed your comment and updated the pull request.

PR #1100, opened 2015-12-15
 ccpp: add AllowedUsers and AllowedGroups feature

mhabrnal:
The feature allows dump core only for allowed users.

The logic is the following:
 - if both options are not-defined or empty keep all core dumps
 - else if crashed UID is in the list of users keep the core dump
 - else if crashed UID belongs to a group in the list of groups keep the core dump

Related to rhbz#1256705

**Requires:** https://github.com/abrt/libreport/pull/398


jfilak:
Thank you!

PR #1102, opened 2015-12-16
testsuite: event-configuration: add missing xml files

martinky82:
They are present on master but not on rhel7 branch, causing this test to
fail all the time.


trapas:

PR #1111, opened 2016-02-16
a-a-save-package-data: blacklist /usr/lib(64)/firefox/plugin-container

mhabrnal:
/usr/lib(64)/firefox/plugin-container is a sandbox/launcher for Firefox
plug-ins. When it crases Firefox loads it again so we don't want those crashes
reported.

Related to: rhbz#1308840


jfilak:
Thank you!

PR #1114, opened 2016-02-23
Problems2 D-Bus API

jfilak:


jfilak:
I have to change the tests a bit as our testing machines cannot create 2GiB files.

src/lib/notify_new_path.c
```diff
@@ -34,18 +42,68 @@ void notify_new_path(const char *path)
...
      * aka "good practice".
      */
     shutdown(fd, SHUT_WR);
+    if (message == NULL)
+    {
+        close(fd);
+        return 0;
+    }
+
+    *message = NULL;
```
kdudka: Looks like a redundant statement to me.

jfilak:
I reduced disk space consumption to hundreds of megabytes from gigabytes

src/lib/notify_new_path.c
jfilak: Fixed, thanks!

mhabrnal:
Otherwise looks good to me, test works as well.

jfilak:
I pushed updated commits to master. Tests passes, `make rpm` works.

PR #1117, opened 2016-03-18
RHEL7: vmcore: catch IOErrors and OSErrors

mhabrnal:
Perhaps some temporary data cleaner removed problem directory while the hook
was still using in.

Resolves: rhbz#1077241


mhabrnal:
retest this please

PR #1122, opened 2016-03-30
RHEL7: Save Vendor and GPG Fingerprint + testcases

mhabrnal:
Related to rhbz#1258482
Requires https://github.com/abrt/libreport/pull/422


mhabrnal:
retest this please

PR #1133, opened 2016-04-26
Docker updates

jfilak:


src/daemon/abrt-action-save-container-data.c
```diff
@@ -29,58 +29,109 @@ void dump_docker_info(struct dump_dir *dd, const char *root_dir)
...
+
+        last = strrchr(last, '-');
+        if (last == NULL)
+        {
+            log_debug("The docker mount point has unknown format");
+            continue;
+        }
+
+        ++last;
+        container_id = xstrndup(last, 12);
```
mhabrnal: Could you please use some constant instead of integer 12?

jfilak: Sorry, I don't see any reason for it.  The value is literally a magic constant. I will add a comment.

jfilak:
I've added the comment describing the magic value 12.

mhabrnal:
Looks good to me.

PR #1136, opened 2016-04-27
a-a-install-debuginfo: correct handling of DebuginfoLocation

### LINKED ISSUE #1135: abrt-action-install-debuginfo deletes files from "/"
I had something crash on me the other day and I went to generate a backtrace with abrt. It seemed to take longer than usual, so I worked on something else and maybe 10 minutes later, once I noticed the machine felt a little sluggish, I came back to look at what abrt was doing.  

I found it was deleting files from **"/"** -- a fairly significant negative side effect that was completely unexpected (I mean really ... attempting to submit a bug report suddenly causes files to be deleted from the root file system?!).

Here is the output from `ps -o user,cmd | fgrep abrt-action-trim-files`  showing the process doing the actual deletion:

> abrt     abrt-action-trim-files -f 4096m:/ -- //usr/lib/debug/.build-id/17/87e030bb7c3878dfda2f910a4ccc56c9deef18.debug ... ... ... /path/to/many/.debug

you'll notice the **-f 4096m:/** argument to abrt-action-trim-files, that **"/"** is supposed to be the first debug info cache directory in a list of debuginfo cache directories.  Clearly doing an **"rm -rf /"** is not the desired behavior.

Unfortunately, when the config option key, **DebuginfoLocation**, is set in _CCpp.conf_, then lines 146-152 of `abrt-action-install-debuginfo` end up initializing `cachedirs` to a string value "/var/cache/abrt-di" instead of the list type ["/var/cache/abrt-di"]:
`conf = problem.load_plugin_conf_file("CCpp.conf")`
`cachedirs = conf.get("DebuginfoLocation", None)`
Then, later on line 189, when building the command line arguments to `abrt-action-trim-files`, the code takes the first index of `cachedirs`.  When `cachedirs` is a string type whose value is an absolute directory path, taking the first index(`cachedirs[0]`), yields **"/"** instead of the first directory entry in a list:
`argv = ["abrt-action-trim-files", "-f", "%um:%s" % (size_mb, cachedirs[0]), "--"]`

So, not only is `abrt-action-trim-files` instructed to delete from **"/"**, but it is also only given a single cache directory to clean up, when supposedly there can be multiple cache directories specified -- that is the whole reason `cachedirs` is supposed to be a list in the first place!

jfilak:
Closes #1135

### DIFF
--- a/src/hooks/CCpp.conf
+++ b/src/hooks/CCpp.conf
@@ -34,7 +34,15 @@ SaveFullCore = yes
 # Used for debugging the hook
 #VerboseLog = 2
 
-# Specify where you want to store debuginfos (default: /var/cache/abrt-di)
+# Specify directories where ABRT should look for non-system debuginfos.
+#
+# Add a colon separated list of file system paths.
+#
+# Beware the first path in the list is used by ABRT to save downloaded
+# debuginfos, therefore the first path in the list must be _writable_, the
+# rest of the list can be read-only.
+#
+# (default: /var/cache/abrt-di)
 #
 #DebuginfoLocation = /var/cache/abrt-di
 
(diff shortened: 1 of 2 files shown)


mhabrnal:
Looks good to me.

PR #1140, opened 2016-05-02
RHEL7: Fix memory leaks in abrt-dbus

jfilak:
Fix several repeated leaks that were causing abrt-dbus to waste system
memory.

I used this valgrind command:
    valgrind --tool=memcheck --leak-check=full --show-leak-kinds=all \
             --track-origins=yes --suppressions=glib.supp \
             --log-file=/tmp/leaks-$(date +%s).txt abrt-dbus -vvv -t 10

With suppressions from libsecret and NetworkManager:
  * https://raw.githubusercontent.com/GNOME/libsecret/master/build/glib.supp
  * https://raw.githubusercontent.com/NetworkManager/NetworkManager/master/valgrind.suppressions

The suppressions were needed because Glib allocates a lot of static
stuff and does not free it at exit because it is useless.

Resolves: #1319704

### DIFF
--- a/src/dbus/abrt-dbus.c
+++ b/src/dbus/abrt-dbus.c
@@ -97,22 +97,21 @@ static uid_t get_caller_uid(GDBusConnection *connection, GDBusMethodInvocation *
     GError *error = NULL;
     guint caller_uid;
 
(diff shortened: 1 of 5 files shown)


mhabrnal:
Looks good to me.

PR #1142, opened 2016-05-02
testsuite: add test for RequireAbsolutePath option

mhabrnal:
The option is placed in '/etc/abrt/plugins/python.conf' and
the option is placed in '/etc/abrt/plugins/python3.conf'.

### DIFF
--- a/tests/runtests/python-addon/runtest.sh
+++ b/tests/runtests/python-addon/runtest.sh
@@ -47,6 +47,8 @@ rlJournalStart
     rlPhaseStartTest
         rlRun "sed -i '2i 0/0 # error line' $TFILE" 0 "Add error line to $TFILE"
         cat $TFILE
+
+        prepare
         rlRun "python $TFILE" 1 "Run python $TFILE"
         sleep 3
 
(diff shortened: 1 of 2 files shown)


tests/runtests/python-addon/runtest.sh
```diff
@@ -57,6 +59,37 @@ rlJournalStart
...
+1/0
+EOF
+        rlRun "chmod +x $CRASH_PY" 0
+
+        rlRun "augtool set /files${AASPD_CONF}/ProcessUnpackaged yes" 0
+        rlRun "augtool set /files${PYTHON_CONF}/RequireAbsolutePath yes" 0
+
+        prepare
+        rlRun "./$CRASH_PY" 1
+        rlAssertNotExists $crash_PATH
```
jfilak: This asserts test if `$crash_PATH` does not exist but if the crash does not exist, `$crash_PATH` is empty. What's worse, `$crash_PATH` is populated by the previous test which also removes the related problem, so the variable is not empty and the path will never exist.

tests/runtests/python-addon/runtest.sh
```diff
@@ -57,6 +59,37 @@ rlJournalStart
...
+
+        rlRun "augtool set /files${PYTHON_CONF}/RequireAbsolutePath no" 0
+
+        prepare
+        rlRun "./$CRASH_PY" 1
+        wait_for_hooks
+        get_crash_path
+        check_dump_dir_attributes $crash_PATH
```
jfilak: What if abrt python handler causes python to crash?

mhabrnal: I've added a check if ABRT caught the right crash.

mhabrnal:
Updated

jfilak:
Thank you! Good job.

PR #1144, opened 2016-05-17
systemd-journal core dump file fixes

jfilak:
Without this pull request, abrt-action-dump-journal-core saves coredump files uncompressed, so the problem directories are unusable.


mhabrnal:
Looks good to me

PR #1151, opened 2016-06-08
testsuite: mailx-reporting - hardcode locale and timezone

martinky82:
The test is failing if there is different locale and timezone set than on the system where the templates were generated. Thus, I'm hardcoding it to use LANG=C and timezone Europe/Prague so the date / time is generated in the same format as expected.


trapas:

mhabrnal:
Thank you!

PR #1154, opened 2016-06-13
Vmcore partitions

jfilak:


mhabrnal:
Manually tested, looks good to me.

PR #1155, opened 2016-06-16
Ultimately fix the problems with finding moutpoint in the vmcore harvestor

jfilak:
findmnt solves both problems:
* UUID=/LABEL=
* different device links

Related: rhbz#1147053


mhabrnal:
Tested, looks good to me.

PR #1160, opened 2016-07-12
Allow selinux to be optional at build time

wlindauer:


trapas:

jfilak:
Thank you!

PR #1161, opened 2016-07-13
abrt-hook-ccpp: Fix mismatching argument

wlindauer:
The calling code in abrt-install-ccpp-hook was passing in --test-config.
This was only working due to the misuse of strcmp, so it was accepting
any string other than --config-test.


trapas:

jfilak:
Good catch. Thank you!

PR #1163, opened 2016-07-21
hook-ccpp: dump own core file in debug mode

jfilak:
Kernel sets RLIMI_CORE to 1 for core_pattern processes in order to avoid
crash recursion. Thus if we want to be able to dump own corefile, we
must set RLIMIT_CORE !=1.

In order to avoid unintended disasters cause by crash recursion, the
core file will be produced only if DebugLevel is >= 100. This
should ensure that we can debug abrt problems without infinite recursion
in abrt-hook-ccpp.

SELinux prevents abrt-hook-ccpp from changing its own RLIMIT_CORE,
therefore, SELinux must be in permissive mode at least. Otherwise, the
core file won't be produced as RLIMIT_CORE will remain set to 1.

### DIFF
--- a/src/hooks/abrt-hook-ccpp.c
+++ b/src/hooks/abrt-hook-ccpp.c
@@ -27,6 +27,8 @@
 typedef char *security_context_t;
 #endif
 
+#include <sys/resource.h>
+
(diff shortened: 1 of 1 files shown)


mhabrnal:
Looks good to me, tested.

Issue #1166, opened 2016-07-28
[abrt on fedora] ABRT completely refuses to assist in reporting issues on many essential applications due to "locked memory"


JonasT:
ABRT completely refuses to assist in reporting issues on many essential applications due to "locked memory":

This means automatic reporting issues on central applications like gnome-shell (central window manager of the popular GNOME and also the even more central compositor for GNOME Wayland) or evolution (main e-mail client of GNOME) is completely disabled. This is even more problematic because both of those applications are somewhat common sporadic crashers due to their complexity, so it's hard to run them on gdb for a specific issue because it's usually not an easy-to-reproduce bug and it's on the other hand still quite necessary to report crashes for them because they do like to have sporadic weird crashes and they are such central programs to the linux desktop.

Because not even local backtrace generation with debug symbols download is permitted by ABRT in such a case, there's pretty much no way to report any of the problems unless I invest a lot of energy reading up on how to best use gdb, and possibly wasting a lot of time trying to reproduce a sporadic crash again because ABRT refused to do anything with the already observed crash.

(also while I might be able to do that as someone with C/C++ developer experience if I invest some time - which also possibly not everyone can afford to do - the average user certainly couldn't)

As I understand it, locked memory areas with potentially protected passwords will most likely show up in a crash dump, and maybe show up in a backtrace. However, you already ask the user to review the backtrace if generated locally. Would it therefore be possible to only disable crash dump uploads for such programs with locked memory areas with a warning, and still allow reviewing and submitting a backtrace if generated locally?

xsuchy:
@mhabrnal can you please update this issue? Is it done, or needs some work?

jfilak:
Not yet fixed in gnome-abrt.

PR #1170, opened 2016-08-03
Introduce unsafe reporting for not-reporable problems

mhabrnal:
Parameter unsafe ignores security checks and allows to report not reportable problems.

Related to #1351297
Related to #1166

### DIFF
--- a/src/cli/process.c
+++ b/src/cli/process.c
@@ -32,7 +32,7 @@ enum {
     ACT_SKIP
 };
 
-static int process_one_crash(problem_data_t *problem_data)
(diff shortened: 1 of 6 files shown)


src/cli/process.c
```diff
@@ -32,7 +32,7 @@ enum {
     ACT_SKIP
 };
 
-static int process_one_crash(problem_data_t *problem_data)
+static int process_one_crash(problem_data_t *problem_data, int flags)
```
jfilak: If you want to reuse CMD_REPORT_* members (definitely good idea), you should name the parameter "cmd_flags" because I can imagine similar flags for the status and info commands.

PR #1172, opened 2016-08-04
abrt-bodhi: use CCpp PackageManager configuration directive from conf…

arhledvink:
resolves #bz1357811
issue tracked as (no idea)
adds configure option
alters package manager directive in config file
suggests command with both package manager and previously omited repository
the test suite query is from live bodhi. not sure this is the best idea but...


trapas:

jfilak:
add to whitelist

jfilak:
test this please

jfilak:
Thank you! I pushed commit https://github.com/abrt/abrt/commit/6858060e4d3ceb20464948967e87deb22b0d95da

PR #1181, opened 2016-09-17
Headless gdb

jfilak:
Use /usr/libexec/gdb for non-interactive tasks.


jfilak:
The build failed because gdb-headless is not on Fedora24 and will not be there.

PR #1184, opened 2016-10-17
Dump core files faster and limit their size

jfilak:


mhabrnal:
The pull request is ready to be merged as soon as the pointers comparison issues are fixed.

tests/runtests/ccpp-plugin-core-size/runtest.sh
```diff
@@ -0,0 +1,215 @@
...
+    rlPhaseStartSetup
+        check_prior_crashes
+        load_abrt_conf
+
+        TmpDir=$(mktemp -d)
+        rlRun "gcc -Wall -std=gnu99 -pedantic -o $TmpDir/$CRASHER $CRASHER.c"
+        pushd $TmpDir
+
+        rlFileBackup $ABRT_CONF $CCPP_CONF $AASPD_CONF
+        rlRun "augtool set /files${AASPD_CONF}/ProcessUnpackaged yes"
```
mhabrnal: Make sure the MakeCompatCore option is allowed in CCpp.conf as well.

mhabrnal:
Looks good to me!

mkutlak:
> did this make it into rhel7? I can't seem to find these options in my CCpp.conf

No, this change did not make it to RHEL7.

Klaas-:
this means abrt-ccpp under rhel7 will not honor MaxCrashReportsSize and potentially create huge core files in /var, right?

xsuchy:
@Klaas- We only backport code which has big severity or security impact or which is a request by our customers. This is none of these. If you want to have this backported to RHEL7 then I suggest you contact our support and open a case.

PR #1195, opened 2016-11-01
doc: we need gettext-devel to be able to run autogen.sh

xsuchy:
This is need to run autopoint in autogen.sh. It is not needed in RPM itself, so it is not listed as build requires.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -18,7 +18,7 @@ EOH
 
 build_depslist()
 {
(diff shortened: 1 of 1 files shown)


trapas:

jfilak:
What about to put install gettext-devel in `./autogen.sh sysdpes`?

xsuchy:
updated

jfilak:
What do you think about this commit:
https://github.com/jfilak/abrt/commit/efa734e3e782a8d0ef759073b9a6c4d92268ffc8

jfilak:
I pushed: https://github.com/abrt/abrt/commit/efa734e3e782a8d0ef759073b9a6c4d92268ffc8

PR #1200, opened 2016-11-09
Bodhi event

jfilak:
Analyzing of C/C++ problems should not always include finding Bodhi update - for example reporting a problem via Email does not require that. I know there were some attempts to sort out this issue in the history and I was refusing them but now I came up with an acceptable solution.

This PR introduces a new event call 'analyze_BodhiUpdates' and abrt/libreport#464 adds the event to Fedora workflows. This seems to me as the most natural approach. This approach will also enable searching for Bodhi updates for other problem types than just C/C++ problems.


mhabrnal:
Looks good to me, tested.

PR #1203, opened 2016-11-11
reporter-systemd-journal improvements and minor fixes

mhabrnal:


src/plugins/abrt-action-notify
```diff
@@ -111,7 +111,7 @@ def emit_crash_dbus_signal(problem_data):
 
         # List of tuples where the first member is element name and the second
         # member is a Boolean flag which is True if the element is required
-        arguments = ((FILENAME_PACKAGE, True), (CD_DUMPDIR, True),
+        arguments = ((FILENAME_PACKAGE, False), (CD_DUMPDIR, True),
```
jfilak: Just for the record I was concerned about `abrt-applet` [1] but the code below adds empty string for missing arguments.

1: https://github.com/abrt/abrt/blob/master/src/applet/applet.c#L999

jfilak:
Good job! Thank you.

PR #1206, opened 2016-11-21
Import gobject from gi.repository if gi was already imported

vojtechtrefny:
It isn't possible to 'import gobject' if gi/gi.repository was
imported before.

---
```AttributeError: When using gi.repository you must not import static modules like "gobject". Please change all occurrences of "import gobject" to "from gi.repository import GObject". See: https://bugzilla.gnome.org/show_bug.cgi?id=709183```

This is a problem when trying to use `ProblemWatcher` from applications using `gi.repository` (so for example all Gtk applications).

### DIFF
--- a/src/python-problem/problem/watch.py
+++ b/src/python-problem/problem/watch.py
@@ -15,7 +15,7 @@ class ProblemWatcher(object):
 
     def __init__(self, auth):
         import dbus
-        import gobject
+        from gi.repository import GObject as gobject
         from dbus.mainloop.glib import DBusGMainLoop
 
         gobject.threads_init()

### CODE REVIEW

**Comment by trapas** [comment]

File: src/python-problem/problem/watch.py [hunk 1]
```diff
@@ -15,9 +16,13 @@ class ProblemWatcher(object):
...
     def __init__(self, auth):
         import dbus
-        import gobject
         from dbus.mainloop.glib import DBusGMainLoop
 
+        if 'gi' in sys.modules.keys():
+            from gi.repository import GObject as gobject
+        else:
+            import gobject
+
```
jfilak: Do we really need to this exercise or can we just replace `import gobject` with `from gi.repository import GObject as gobject`?

vojtechtrefny: Fixed. I tried to keep backwards compatibility with pygobject2 but if it isn't a problem, it's ok to use just the `gi.repository`.

jfilak:
Well done! Thank you!

PR #1207, opened 2016-11-23
Ignore not reportable

jfilak:


jfilak:
> There is no log message in case reporting via cli-ng cannot be done because the problem is not reportable.

The problem is that there are two execution paths and I was testing my code on the path where libreport prints out a message. If I had known that libreport ignores the 'not-reportable' check fro events that has 'skip review' set to yes, I would have care about it.

> I would like cli-ng have the same behavior as abrt-cli. If problem is not reportable, log messages with reason is printed and no workflow is presented.

In my test case, where the first executed event has 'skip review` set to no, there were a message printed out by libreport immediately after a workflow was selected. Thus, I didn't find it necessary to add an extra check to abrt.

PR #1209, opened 2016-11-28
python: create analyzer element in dumpdir

mhabrnal:
Stop creating of an analyzer element was introduced in commit
61b879a8efa04a41f5f75f7b0ef0fbe4707a42c0.
We still want to create the element for Python and Python3.

For Python problems the analyzer is abrt-python-handler and for Python3
problems the analyzer is abrt-python3-handler.


jfilak:
Thank you! I pushed commit a153476a88b78abe0fdd228d82734d5ee52d929c

PR #1210, opened 2016-11-29
ccpp: tell gdb to analyze saved binary image

jfilak:
This is particularly useful when dealing with crashes in containers and
changed root environments. GDB reads the debug information from the
source elf and if the source elf is not available GDB generates useless
backtraces.


mhabrnal:
Looks good to me.

PR #1211, opened 2016-11-29
cli-ng: fix abrt list --fmt parameter

mhabrnal:
Argument of parameter --fmt was not used even it is defined.

$ abrt list {what} raised exception UnboundLocalError.


jfilak:
Thank you. I pushed commit e7558f70d433df4ae1c4cb88e824e8284631efc0

PR #1215, opened 2016-12-20
testsuite: a-i-event-loop - sleep to avoid race cond

martinky82:
set ProcessUnpackaged to yes


trapas:

martinky82:
Please, drop the chunk waiting for sos to finish and replace it with plain 5s sleep.

mhabrnal:
Thank you!

PR #1216, opened 2016-12-28
testsuite: add test for bugzilla-default-url

jcpunk:
Adds test requested in https://github.com/abrt/libreport/pull/469


trapas:

jfilak:
Well, you don't even need to update `/etc/hosts` as you can just set BUG_REPORT_URL to 127.0.0.1 in the os_info file.

I would also like to ask you to cover more possible situations like BUR_REPORT_URL not found or not set, etc.

jcpunk:
I've added several more tests based on your input.

jfilak:
Thank you! I pushed commit https://github.com/abrt/abrt/commit/c40cb3dac3208069a7a9627343d82451d7ea342c

PR #1218, opened 2017-01-03
daemon: Allow rpm to be optional at build time

wlindauer:
This request is similar to the one I made for satyr, but this time the default should be yes.


trapas:

jfilak:
ok to test

PR #1224, opened 2017-02-10
spec: add rpm-build as BuildRequires

marusak:
Without this patch 'make rpm' command fails with message
`rpmbuild: command not found`.

Signed-off-by: Matej Marusak <mmarusak@redhat.com>

### DIFF
--- a/abrt.spec.in
+++ b/abrt.spec.in
@@ -75,6 +75,7 @@ BuildRequires: %{dbus_devel}
 BuildRequires: gtk3-devel
 BuildRequires: glib2-devel >= 2.43
 BuildRequires: rpm-devel >= 4.6
+BuildRequires: rpm-build
 BuildRequires: desktop-file-utils
 BuildRequires: libnotify-devel
 #why? BuildRequires: file-devel


trapas:

xsuchy:
-1
BuildRequires are dependencies which you need when you have spec file and source. Not the dependencies which you need to create some tool inside this project. This is not needed for actual SRPM rebuild.

xsuchy:
list_build_dependencies() in autogen.sh is more suitable for this.

marusak:
Now it makes sense not to state it as a BuildRequires. But I think it should be listed when 'autogen.sh sysdeps' is run. list_build_dependencies() looks good. I am not sure what does BUILD_SYSTEM_DEPS_LIST exactly mean and if I can add rpm-build to this variable, or should create another one.

xsuchy:
Just add it to that one.

jfilak:
I haven't added _rpm-build_ to the list of dependencies intentionally as it is not required to **build** the project. I believe that not everyone wants to get _rpm-build_ installed when building ABRT from sources.

jfilak:
I like this PR: _rpm-build_ would parse the spec file to find out that it requires a package shipping itself installed on the machine. "_Doing computers_" is so much fun :D If it failed it would die while laughing.

marusak:
Ok, since the `rpm-build` was left out intentionally, I am closing this PR.

Issue #1225, opened 2017-02-22
don't use hawkey


ignatenkobrain:
Hawkey C API is considered unstable and package is going to be removed from Fedora.

I've seen that you use hy_split_nevra() only, feel free to implement your own, it's very easy.

xsuchy:
@ignatenkobrain Any ETA when hawkey will be removed?

ignatenkobrain:
@xsuchy before branching (i.e. less than one week)

PR #1227, opened 2017-02-27
Remove dependency on hawkey, replace by own implementation

juliusmilan:
Function hy_split_nevra of obsolete hawkey library replaced.
Issue #1225


trapas:

PR #1228, opened 2017-02-27
Remove dependency on obsolete hawkey library

### LINKED ISSUE #1225: don't use hawkey

juliusmilan:
Resolves #1225


trapas:

mhabrnal:
test this please

src/plugins/bodhi.c
```diff
@@ -271,6 +270,36 @@ static void bodhi_print_errors_from_json(json_object *json)
     return;
 }
 
+/**
+ * Parses only name of rpm package
+ */
+static int parse_pkginfo_name(const char *pkg_full_name, char **name)
```
jfilak: How this function works if you pass a name that does not match package naming rules? It would be super awesome if it does not crash.

jfilak:
I believe that the parser function deserves some testing.

src/plugins/bodhi.c
juliusmilan: Yes it dies, but it keeps the same behavior as with original hawkey function. When name doesn't match package naming rules, function return EINVAL (invalid input). I just replaced library function with its simplified equivalent. What do you suggest? Should it just log message and continue in cycle?

jfilak: Oh dear, you are right. I've missed the `c <= pkg_full_name` test. Sorry! Please, ignore my initial comment.

However, I am wondering what format of package name the function should parse. Is it ```foo-1.2.3-4.fc26.x86_64``` or without the arch part?

jfilak:
Yeah, static functions are hard to test. We use mock bodhi responses in this case:
https://github.com/abrt/abrt/tree/master/tests/runtests/bodhi

src/plugins/bodhi.c
juliusmilan: with arch part, I will put the comments there, because it is not clear, thank you.

jfilak: Oh, I see, the function can only parse the name element. The function says it but I didn't get it. How about to add a short description of the algorithm like: "We reverse iterate through the string. When we go over the first dot we parsed release and we can start searching for - that delimits Version from Release. Once we find the Version Release delimiter we can start looking for Name Version delimiter which is again -."

I'm not sure why we need to search for the last dot. It would be nice if you could add a few examples of package names. Or what about to add a link to package name format specification?

juliusmilan: Sure, I will write explaining comments.
But I found some mistake, RPM naming convention is:
name-version-release.architecture -> ("easy-1.2.3-4.fc18.x86_64")
this format is parsed (hawkey & our new code) but it seems we used it badly until now, because we get "nvr" just in format:
name-version-release -> ("dnf-plugins-core-1000.1.21-2.fc24" in runtests/bodhi/queries/pkgmgr_query)
so our parsed values "were shifted by one", ending with parsed name shorter or returned error.
The only question now is if change of this will not affect existing running solution somewhere in production, what do you think?

jfilak: Good catch! I have added the parse `nevra` function while I should have add a parse `nvr` function. How comes the tests didn't reveal this issue? Can you please analyze it more and update the tests to do better job?

I don's see any problem in fixing this issue and I don't believe it can cause any problems as the only production is Fedora Bodhi.

juliusmilan:
added comments,
parsing function corrected/modified to parse nvr (name-version-release) format as appropriate
added tests covering new function

src/plugins/bodhi.c
juliusmilan: Thank you, There were no tests covering this functionality i.e. parsed values returned by hy_split_nevra. I added asserts into tests to check parsed names.
Great, so then I fixed the function.

mhabrnal:
I like the pull request. I would rather see the test more complex like add some non standard or bad inputs and so on but I thing is not necessary in this case. If @jfilak agree, I will merge the PR.

j-mracek:
I was requested if function hy_split_nevra is present in libdnf and according to my search answer is yes, it is present in hy-util.h , but don't know the plans. So far dnf use it through python binding in hawkey.split_nevra(s) function.

jfilak:
@j-mracek We're not going to add a dependency on DNF again. This approach proved wrong ;) Thank you! If there were such a function in RPM library, I would go for it.

jfilak:
@mhabrnal Agreed. But who will add the missing tests and when?

mhabrnal:
@jfilak the test probably won't be improved in the future if the testcase is not improved in the pull request. I wanted to say in this case I am fine with proposed test. But you are right, all should be tested properly.
@juliusmilan can you please improve the test?

juliusmilan:
@mhabrnal yes sure, I am going to add the tests

mhabrnal:
Looks good to me. Good work!

Issue #1033, opened 2015-08-25
Do not send MCEs as uReports


mbrysa:
All MCEs are invalid uReports because they don't contain a backtrace.
They are linked to the hardware where the problem occurred, but there's no way to identify it so they are worthless from a statistics point of view as well.

sorki:
Related to #889

PR #1234, opened 2017-03-16
koops_event: add check to not send MCEs as uReports

juliusmilan:
Resolves abrt/abrt#1033

I think the best place to check for attempt to send MCE ureport is in
src/plugins/koops_event.conf -> EVENT=report_uReport
it would be possible in abrt-action-ureport too, but abrt-action-ureport is used by other scripts as
(ccpp_event.conf, python3_event.conf, ...) and those have nothing to do with MCEs

### DIFF
--- a/src/plugins/koops_event.conf
+++ b/src/plugins/koops_event.conf
@@ -36,14 +36,18 @@ EVENT=report_Bugzilla type=Kerneloops
 
 # Send micro report
 EVENT=report_uReport type=Kerneloops
-        /usr/libexec/abrt-action-ureport
+        if [ ! -e mce ]; then
+            /usr/libexec/abrt-action-ureport
+        else
+            echo "Not reportable, problem has hardware character (MCE)"
+        fi
 
 # Update ABRT database after successful report to bugzilla
 EVENT=post_report type=Kerneloops
         reporter-ureport -A -B
 
 # Automatic/simple GUI-based kernel oopses reporting will do this:
-EVENT=report-gui type=Kerneloops
+EVENT=report-gui type=Kerneloops mce!=non-fatal
         report-gtk -- "$DUMP_DIR"
 
 EVENT=report-cli type=Kerneloops
--- a/src/plugins/abrt-action-check-oops-for-hw-error.in
+++ b/src/plugins/abrt-action-check-oops-for-hw-error.in
@@ -108,6 +108,9 @@ if __name__ == "__main__":
         with open_or_die("kernel", "w") as krnlfile:
             krnlfile.write(kernel)
 
+    with open_or_die("mce", "w") as f:
+        f.write("fatal" if vmcore_mce else "non-fatal")
+
     # vmcore MCEs already have good backtrace element, nothing more to do
     if vmcore_mce:
         sys.exit(0)


trapas:

mhabrnal:
test this please

jfilak:
What about to change the type of problem in https://github.com/abrt/abrt/blob/master/src/plugins/abrt-action-check-oops-for-hw-error.in instead? We don't need to use any kind of workaround here because we own the code :)

juliusmilan:
Great idea! thank you :) I am just wondering what all kinds of events will be needed after this change.
Do we want to inform user (report-gui)? I think no, because of #1033 but maybe we want to inform user about fatal MCEs. But no other events comes to my mind.

jfilak:
Well, we can go crazy and move support of "MCE" from abrt-dump-*-oops to a completely new tool detecting just MCEs. We could drop `abrt-action-check-for-hw-oops` then. We probably need a generic journald watcher, so we don't need to introduce a new service when we come to need to detect a new type of problems in journald.

See https://github.com/abrt/abrt/blob/master/src/lib/kernel.c#L133

jfilak:
What about to make it simple. Just create a new element in https://github.com/abrt/abrt/blob/master/src/plugins/abrt-action-check-oops-for-hw-error.in and stop reporting if the new element exists. It will be implemented in a minute and will not be such a hack. However, if you have plenty of time, I encourage you to implement this idea: https://github.com/abrt/abrt/pull/1234#issuecomment-287111212

juliusmilan:
I updated it, picked the simple way, but I will keep in mind the idea of specialized MCE watcher.

jfilak:
Thanks! It looks good but would be nice if you could describe your ideas in Commit message.
Why you decided for this approach, why you treat 'fatal' and 'non-fatal' different way (and how), future plans and so on.

juliusmilan:
Yes, sure.
I am just not sure if fatal MCEs should be reported to bugzilla. I think yes, because they are caused by kernel panic and that's I think what kernel developers want to know about.

jfilak:
An MCE is a hardware error, hence, there is only a little chance such a Bugzilla bug can be resolved by Kernel maintainers. If a user really wants to report a not-reportable problem, he/she can do it from command line using '--unsafe' parameter.

juliusmilan:
Thank you for good explanation comment. I copied part of it to the commit message.
And of course disabled Bugzilla reporting of both fatal and non-fatal MCEs.

src/plugins/koops_event.conf
```diff
@@ -30,23 +30,23 @@ EVENT=post-create type=Kerneloops remote!=1
 #EVENT=report_Kerneloops type=Kerneloops
         reporter-kerneloops
 
-EVENT=report_Bugzilla type=Kerneloops
+EVENT=report_Bugzilla type=Kerneloops mce!=non-fatal mce!=fatal
```
jfilak: You don't need to do this. The problem directory already contains 'not-reportable' which prevents users from reporting the problem to Bugzilla without pass '--unsafe'.

src/plugins/koops_event.conf
```diff
@@ -30,23 +30,23 @@ EVENT=post-create type=Kerneloops remote!=1
...
         reporter-bugzilla -b \
                 -F /etc/libreport/plugins/bugzilla_format_kernel.conf
 
 # Send micro report
-EVENT=report_uReport type=Kerneloops
+EVENT=report_uReport type=Kerneloops mce!=non-fatal mce!=fatal
```
jfilak: Have you tested this? Does that produce a strange log messages if an MCE is detected?

src/plugins/koops_event.conf
```diff
@@ -30,23 +30,23 @@ EVENT=post-create type=Kerneloops remote!=1
...
 EVENT=post_report type=Kerneloops
         reporter-ureport -A -B
 
 # Automatic/simple GUI-based kernel oopses reporting will do this:
-EVENT=report-gui type=Kerneloops
+EVENT=report-gui type=Kerneloops mce!=non-fatal
         report-gtk -- "$DUMP_DIR"
 
-EVENT=report-cli type=Kerneloops
+EVENT=report-cli type=Kerneloops mce!=non-fatal
```
jfilak: This effectively prevents users from running 'abrt-cli report --unsafe'.

src/plugins/koops_event.conf
juliusmilan: done

src/plugins/koops_event.conf
juliusmilan: done

src/plugins/koops_event.conf
juliusmilan: done, removed.

src/plugins/koops_event.conf
jfilak: This behavior is expected but I was referring to system logs that appear in journald after a new MCE is detected. In such a case, ABRT runs `abrt-action-ureport` from `abrt-action-notify` (EVENT=notify) and this can cause unnecessary error log messages. Thinking about it a bit more, `abrt-action-notify` may ignore `EVENT=report_uReport` declaration and execute `abrt-action-ureport` or `reporter-ureport` directly.

Could you please double-check that there are no error logs in journald after a new MCE is detected and `abrt-action-notify` does not ignore your configuration changes (ie. ARBT does not send the MCE problem to FAF)?

juliusmilan: Thank you for next insight.
I tried to report using notify event, it respects new settings and doesn't send uReport

Regarding error logs, I think it is ok, this is all I get:
Mar 22 17:21:56 localhost.localdomain abrt-dump-journal-oops[866]: abrt-dump-journal-oops: Found oopses: 1                                                                                                         
Mar 22 17:21:56 localhost.localdomain abrt-dump-journal-oops[866]: abrt-dump-journal-oops: Creating problem directories
Mar 22 17:21:57 localhost.localdomain abrt-dump-journal-oops[866]: Reported 1 kernel oopses to Abrt
Mar 22 17:21:59 localhost.localdomain abrt-server[1068]: Cannot notify '/var/spool/abrt/oops-2017-03-22-17:21:56-866-0' via uReport: No processing is specified for event 'report_uReport'

jfilak: Well, that is not OK. The last message would cause many administrators to start pondering what is wrong and be sure you would get a bug report sooner or later.

The fix is easy -> don't disable the "EVENT=notify", but add 'if [ -e mce ]; then ...; else ...; fi`. If you want to be really cool, you can print out an informative message - but don't forget that that message must not scary administrators - it must be something positive.

juliusmilan: done

jfilak: Good, that's the code what I had on mind, but could you please move it here (`src/plugins/koops_event.conf`)?`EVENT=notify` should stay as it is now, because `abrt-applet` listens for this event and shows the desktop notifications - we definitely want to inform users about new MCEs through desktop notifications.

jfilak: Another option would be to turn of the error message in `abrt-action-notify`. We can say that if there is no AutoReportEvent (`/etc/abrt.conf`) enabled for the notified problem we skip this notification channel. The error message should be printed out in debug mode.

juliusmilan: moved condition to src/plugins/koops_event.conf

jfilak:
Good job! Thank you very much indeed!

Now, FAF will not need to handle hundreds of thousands of invalid uReport a day. I feel a bit embarrassed as I haven't been able to fix this issue sooner.

PR #1237, opened 2017-03-27
reorder function declaration

xsuchy:
addressing:
```
DEBUG: notify_new_path.c: In function 'abrt_notify_new_path':
DEBUG: notify_new_path.c:24:5: error: implicit declaration of function 'notify_new_path_with_response' [-Werror=implicit-function-declaration]
DEBUG:      notify_new_path_with_response(path, NULL);
DEBUG:      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/lib/notify_new_path.c
+++ b/src/lib/notify_new_path.c
@@ -18,12 +18,6 @@
 #include <sys/un.h>
 #include "libabrt.h"
 
-void notify_new_path(const char *path)
-{
-    /* Ignore results and don't wait for response -> NULL */
-    notify_new_path_with_response(path, NULL);
-}
-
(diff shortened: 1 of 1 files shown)


jfilak:
Well, you should rather look what is wrong in `libabrt.h` because these two functions are declared there, so the order of declarations should not matter.

jfilak:
I see. I did a typo, an active user has fixed the typo but has forgotten to correct the typo in all places:
https://github.com/abrt/abrt/blob/master/src/include/libabrt.h#L131

xsuchy:
OK. Closing this one.

PR #1239, opened 2017-03-30
tests - cli-sanity: re-enable report non-reportable phase

martinky82:
BZ#1257159 needs to be tested


trapas:

PR #1240, opened 2017-04-10
Fix 1439821

juliusmilan:
Fixes rhbz#1439821.
Problem was, that regular kernel panic message was treated as fatal MCE, for which we don't want to have  backtrace.
The logic in koops_extract_oopses_from_lines is a bit complex, so I describe it:
If fatal MCE was recognized, dumping of backtrace was skipped by following to the next part of cycle (hacky, in the next part of cycle, there is condition with many allowed strings to continue but not "Call Trace"). The check responsible for this "jump" (kernel.c:396) just checked whether there is a string beginning with "Kernel panic - not syncing", however this string appears in all kernel panics, not just fatal MCEs.

Fatal MCE may have form of:
"Kernel panic - not syncing: Fatal Machine check" or
"Kernel panic - not syncing: Machine check" or I found even
"Kernel panic - not syncing: Uncorrected machine check"
so I modified the condition appropriately and added test covering this case.

### DIFF
--- a/src/lib/kernel.c
+++ b/src/lib/kernel.c
@@ -393,7 +393,7 @@ void koops_extract_oopses_from_lines(GList **oops_list, const struct abrt_koops_
              * In order to capture all these lines, we treat final line
              * as "backtrace" (which is admittedly a hack):
              */
-            if (strstr(curline, "Kernel panic - not syncing"))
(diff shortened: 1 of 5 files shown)


jfilak:
Well done! Would be better to move the PR comment to the commit, though.

juliusmilan:
Yes, you are right, commit message changed.

PR #1242, opened 2017-04-11
abrt_event: Save cpuinfo in problem directories

juliusmilan:
added under "remote!=1" condition, because we don't want to save cpuinfo for uploaded problems

tested for all kinds of problems, also gnome-abrt and abrt-cli list looks well, in output of "abrt-cli list --detailed" appears as:
... 
cpuinfo:        Text file, 4632 bytes
...

### DIFF
--- a/src/daemon/abrt_event.conf
+++ b/src/daemon/abrt_event.conf
@@ -63,12 +63,20 @@ EVENT=post-create container_cmdline!= remote!=1
 #EVENT=post-create
         rm uid; chmod a+rX .
 
-
-# uid file is missing for problems visible to all users
-# (oops scanner is often set up to not create it).
-# Record username only if uid element is present:
 EVENT=post-create remote!=1
+        # uid file is missing for problems visible to all users
+        # (oops scanner is often set up to not create it).
+        # Record username only if uid element is present:
         if [ -f uid ]; then getent passwd "`cat uid`" | cut -d: -f1 >username; fi
+        # Save cpuinfo because crashes in some components are
+        # related to HW acceleration. The file must be captured for all crashes
+        # because of the library vs. executable problem.
+        if command -v lscpu >/dev/null 2>&1; then
+            # use lscpu if installed
+            lscpu > $DUMP_DIR/cpuinfo
+        else
+            cp /proc/cpuinfo $DUMP_DIR/cpuinfo
+        fi
 
 # Record runlevel (if not yet done) and don't return non-0 if it fails:
 EVENT=post-create runlevel= remote!=1
--- a/tests/runtests/dumpdir_completedness/runtest.sh
+++ b/tests/runtests/dumpdir_completedness/runtest.sh
@@ -32,7 +32,7 @@
 TEST="dumpdir_completedness"
 PACKAGE="abrt"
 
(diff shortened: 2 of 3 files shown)


juliusmilan:
tested for all kinds of problems,
added under "remote!=1" condition, because we don't want to save cpuinfo for uploaded problems

jfilak:
LGTM.
Don't forget to update the tests:
https://github.com/abrt/abrt/blob/master/tests/runtests/dumpdir_completedness/runtest.sh
https://github.com/abrt/abrt/blob/master/tests/runtests/oops-processing/runtest.sh

Out of curiosity, did you asked anyone for his opinion on publishing CPUINFO in Bugzilla bugs?

juliusmilan:
Oh sure, will add the tests.
Yes, people from kernel and toolset, there was positive reaction, some people wrote it would be definitely useful. The only response that makes me uncertain:
"It could be useful, but it may leak partner confidential information."
I wonder how to decide if cpuinfo is really confidential. If so, do we already have some similar cases? I mean some process for attaching confidential information. I didn't find.

jfilak:
> "It could be useful, but it may leak partner confidential information."

There is a github issue or a bugzilla mentioning that FAF reports including "full" kernel oops can be used to spy the reporter because it might contain Wi-Fi ESSID or something like that.

So, we should not include entire cpuinfo in uRerport.

Ah, I just realized that you should make the 'cpuinfo' problem element editable to allow users to change its contents while reviewing data to before opening a Bugzilla bug:
https://github.com/abrt/libreport/blob/master/src/lib/problem_data.c#L267

xsuchy:
How can /proc/cpuinfo contain sensitive information? Full kernel oops ok, but cpuinfo?

juliusmilan:
I also think that there is nothing secret or confidential in cpuinfo. I will ask them your question: "how it could leak confidential info" and if they provide reasonable arguments I will solve it as Jakub explained i.e. making file editable.

The next proposal from discussion was to append rather lscpu output, because it is better readable than cpuinfo and they told that in some cases contains next information.

juliusmilan:
Updated to use lscpu, as people wanted.
Updated tests to check lscpu_out file for existence

juliusmilan:
Now I realized that maybe I should add lscpu as required to spec file, what do you think?

xsuchy:
@juliusmilan lscpu is included in util-linux. Yes, please add it to requires in spec file

jfilak:
> Now I realized that maybe I should add lscpu as required to spec file, what do you think?

```bash
if which lscpu; then
    lscpu > $DUMP_DIR/cpuinfo
else
    cp /proc/cpuinfo $DUMP_DIR/cpuinfo
fi
```
- use lscpup if avaliable
- save output in `cpuinfo` to have a single element holding CPU details

jfilak:
Don't forget to run your code with SELinux enforcing to be sure SELinux does not prevent ABRT processes from running lscpu.

Do you know the answer for the question about leaking sensitive data?

juliusmilan:
Added check to lscpu existence, however didn't used "which", but rather "command" as explained here:
http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

Checked with SELinux enforcing, works well.

Regarding the question about leaking sensitive data, here is the answer:
> We have machines in Beaker which are not available to the general
> public.  I don't know the terms of our NDAs, though.

This I solved and tested as you mentioned, making file editable, PR is already opened:
https://github.com/abrt/libreport/pull/484

jfilak:
Good job!

PR #1244, opened 2017-04-24
Fix 1444924

juliusmilan:
resolves problem with DEBUG messages considered as BUGs in kernel log.
added koops suspicious string blacklist, with one entry yet - "DEBUG"
added test

### DIFF
--- a/src/lib/kernel.c
+++ b/src/lib/kernel.c
@@ -150,6 +150,32 @@ static const char *const s_koops_suspicious_strings[] = {
     NULL
 };
 
+static const char *const s_koops_suspicious_strings_blacklist[] = {
+    /* "BUG:" and "DEBUG:" overlaps, we don't want to recognize DEBUG messages as BUG */
+    "DEBUG:",
+
+    /* Termination */
+    NULL
+};
+
+static bool suspicious_line(const char *line)
+{
+    const char *const *str = s_koops_suspicious_strings;
+    for ( ; *str; ++str)
+        if (strstr(line, *str))
+            break;
+
+    if (!*str)
+        return false;
+
+    str = s_koops_suspicious_strings_blacklist;
+    for ( ; *str; ++str)
+        if (strstr(line, *str))
+           break;
+
+    return !*str;
+}
+
 void koops_print_suspicious_strings(void)
 {
     koops_print_suspicious_strings_filtered(NULL);
@@ -159,6 +185,15 @@ GList *koops_suspicious_strings_list(void)
 {
     GList *strings = NULL;
     for (const char *const *str = s_koops_suspicious_strings; *str; ++str)
+        strings = g_list_prepend(strings, (gpointer)*str);
+
+    return strings;
+}
+
+GList *koops_suspicious_strings_blacklist(void)
+{
+    GList *strings = NULL;
+    for (const char *const *str = s_koops_suspicious_strings_blacklist; *str; ++str)
         strings = g_list_prepend(strings, (gpointer)*str);
 
     return strings;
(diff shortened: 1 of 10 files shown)


src/lib/kernel.c
```diff
@@ -150,6 +150,37 @@ static const char *const s_koops_suspicious_strings[] = {
...
+            in_suspicious_list = true;
+        }
+    }
+
+    bool in_suspicious_blacklist = false;
+    for (const char *const *str = s_koops_suspicious_strings_blacklist; *str; ++str)
+    {
+        if (strstr(line, *str))
+        {
+            in_suspicious_blacklist = true;
```
jfilak: We can stop break the loop with the first matching blacklist item.

src/plugins/abrt-dump-journal-xorg.c
```diff
@@ -114,6 +114,7 @@ static void watch_journald(abrt_journal_t *journal, const char *dump_location, i
         .decorated_cb = abrt_journal_watch_extract_xorg_crashes,
         .decorated_cb_data = &watch_conf,
         .strings = xorg_strings,
+        .blacklisted_strings = NULL,
```
jfilak: Why do you need to touch the Xorg parsing in commit fixing a bug in the Kernel oops parsing?

src/plugins/abrt-journal.h
```diff
@@ -129,6 +129,7 @@ struct abrt_journal_watch_notify_strings
     abrt_journal_watch_callback decorated_cb;
     void *decorated_cb_data;
     GList *strings;
+    GList *blacklisted_strings;
```
jfilak: Can you please create a separate commit for the changes in abrt-journal API?

src/plugins/abrt-journal.c
```diff
@@ -469,15 +469,16 @@ void abrt_journal_watch_notify_strings(abrt_journal_watch_t *watch, void *data)
...
     GList *cur = conf->strings;
-    while (cur)
-    {
+    for (; cur; cur = g_list_next(cur))
         if (strstr(message, cur->data) != NULL)
             break;
 
-        cur = g_list_next(cur);
-    }
+    GList *blacklist_cur = conf->blacklisted_strings;
```
jfilak: With no white-list match, there is no need to iterate over the black-list items.

src/plugins/abrt-journal.h
juliusmilan: done

src/lib/kernel.c
```diff
@@ -150,6 +150,36 @@ static const char *const s_koops_suspicious_strings[] = {
...
+static bool suspicious_line(const char *line)
+{
+    bool in_suspicious_list = false;
+    for (const char *const *str = s_koops_suspicious_strings; *str; ++str)
+        if (strstr(line, *str))
+        {
+            in_suspicious_list = true;
+            break;
+        }
+
```
jfilak: Just a side note. I believe that fast returns would make the code more readable:
```c
const char *const *str = s_koops_suspicious_strings;
for ( ; *str; ++str)
    if (strstr(line, *str))
        break;

if (!*str)
    return false;

str = s_koops_suspicious_strings_blacklist
for ( ; *str; ++str)
    if (strstr(line, *str))
       break;

return !!*str;
```

src/plugins/abrt-dump-journal-xorg.c
jfilak: No need to change anything. This commit makes sense now - I was just confused.

mhabrnal:
Thank you!

PR #1245, opened 2017-05-18
koops: Improve not-reportable for oopses with taint flags

juliusmilan:
Adds detailed info about taint flags to not-reportable. This
additional info contains description of every taint flag that appeared
in particular oops.
It is either good for user information and also solves the problem when users
opened bugs on abrt from misunderstanding tainted flags.

Related to rhbz#1452160

not-reportable now looks like this (3 lines added in this case):
A kernel problem occurred, but your kernel has been tainted (flags:GW).
Taint flags description:                                             << new
G - Proprietary module has not been loaded.          << new
W - Taint on warning.                                               << new
Kernel maintainers are unable to diagnose tainted reports. Tainted modules: vboxpci,vboxnetadp,vboxnetflt,vboxdrv.

### DIFF
--- a/src/lib/kernel.c
+++ b/src/lib/kernel.c
@@ -748,9 +748,10 @@ static const char *const tnts_long[] = {
     /* B */ "System has hit bad_page.",
     /* C */ "Modules from drivers/staging are loaded.",
     /* D */ "Kernel has oopsed before",
-    /* E */ "Unsigned module has been loaded."
+    /* E */ "Unsigned module has been loaded.",
     /* F */ "Module has been forcibly loaded.",
-    /* G */ "Proprietary module has not been loaded.",
+            /* We don't want to be more descriptive about G flag */
+    /* G */ NULL, /* "Proprietary module has not been loaded." */
     /* H */ NULL,
     /* I */ "Working around severe firmware bug.",
     /* J */ NULL,
@@ -782,7 +783,7 @@ char *kernel_tainted_long(const char *tainted_short)
         {
             const char *const txt = tnts_long[tnt_index];
             if (txt)
-                strbuf_append_strf(tnt_long, "%s\n", txt);
+                strbuf_append_strf(tnt_long, "%c - %s\n", tainted_short[0], txt);
         }
 
         ++tainted_short;
--- a/src/plugins/oops-utils.c
+++ b/src/plugins/oops-utils.c
@@ -241,13 +241,12 @@ void abrt_oops_save_data_in_dump_dir(struct dump_dir *dd, char *oops, const char
 
             char *tnt_long = kernel_tainted_long(tainted_short);
             dd_save_text(dd, FILENAME_TAINTED_LONG, tnt_long);
-            free(tnt_long);
 
             struct strbuf *reason = strbuf_new();
             const char *fmt = _("A kernel problem occurred, but your kernel has been "
-                    "tainted (flags:%s). Kernel maintainers are unable to "
-                    "diagnose tainted reports.");
-            strbuf_append_strf(reason, fmt, tainted_short);
+                    "tainted (flags:%s). Explanation:\n%s"
+                    "Kernel maintainers are unable to diagnose tainted reports.");
+            strbuf_append_strf(reason, fmt, tainted_short, tnt_long);
 
             char *modlist = !proc_modules ? NULL : abrt_oops_list_of_tainted_modules(proc_modules);
             if (modlist)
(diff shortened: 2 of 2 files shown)


jfilak:
> Taint flags description: << new

I think this line is superfluous. What about this:
```
A kernel problem occurred, but your kernel has been tainted. Taint flags:
- G : Proprietary module has not been loaded.
- W : Taint on warning.
Kernel maintainers are unable to diagnose tainted reports. Tainted modules: 
vboxpci,vboxnetadp,vboxnetflt,vboxdrv.
```

However, the  G flag isn't the show stopper here, only the W flag. Maybe, we should not show the G flag.

juliusmilan:
Message modified, yes you are right, that sentence was superfluous.

I would rather keep G flag too, it is not showstopper, but as we are listing tainted flags I would list them all in authentic manner as kernel produces them and also it at least prevents users from complaining "abrt told my kernel is tainted but I didn't load anything" as opening message of the bug this is supposed to fix. :)

juliusmilan:
one second please, don't merge it yet, I found something bad

jfilak:
Well, the G flag and the P flag represent the same bit of the taint mask with different values:
```
0 - G
1 - P
```
Therefore the G means 'kernel is not tainted by proprietary modules'.

juliusmilan:
If you wish, I can omit printing of G argument, furthermore it is just one line change.
But I would really prefer to keep it consistent with the way kernel developers set it up.
I think when kernel log contains both GW as a flags, as here:
CPU: 1 PID: 14620 Comm: kworker/u17:0 Tainted: G        W   --------------   3.10.0-229.1.2.el7.x86_64 #1
IMHO it's kind of lie to tell, there wasn't G :) I don't understand why they even print G, but maybe they have some good reason, or maybe they just like it, in any case I think they wouldn't be happy.

Are you sure you want to omit G?

juliusmilan:
or maybe would be good to print somewhere taint flags in short format GW, and then also in long format but without G.

jfilak:
> or maybe would be good to print also taint flags in short format like "GW", and then also in long format but without G.

Yeah, that's what I was going to propose.

juliusmilan:
done. not-reportable now looks as follows:

A kernel problem occurred, but your kernel has been tainted (flags:GW). Explanation:
W - Taint on warning.
Kernel maintainers are unable to diagnose tainted reports. Tainted modules: vboxpci,vboxnetadp,vboxnetflt,vboxdrv.

jfilak:
By the way, the original reporter of bz1452160 understand the meaning of the taint flags, he just disagrees with the impossibility to report W tainted oopses. Why do you believe this PR will make the situation better?

juliusmilan:
I wrote comment to the bug, where I am explaining to him why it is bad idea and nobody reacted yet, so I assume they are ok with it. Snippet from comment:

'''
It it easy fix to enable reporting oopses with arbitrary tainted flags, however
I examined this deeply and discovered interesting facts.

* as W taint flag means warning has already occured, it is not interesting for reporting W tainted oopses because we only care about 1 occurence, which was probably already filed by abrt, so reporting W flaged oopses wouldn't even help resolving particular problem and so makes no sense. (https://bugzilla.redhat.com/show_bug.cgi?id=786280#c2)
* enabling reporting of W tainted oopses is in direct contradiction with resolution of bug 786280, which prevents flood of useles identical or follow-on reports.

My conclusion of these two facts is not to enable reporting of W tainted oopses.

* Regarding G flag, it occurs only besides other taint flags, without W it makes no sense too.
'''

But idea of enabling W flag was introduced later. So I decided to focus on bug opening message which says that abrt says about "non-tainted" kernel that it is tainted, next snippet:

'''
In opening message is written:
"In abrt oops data, the kernel is reports as tainted, though this is a fresh 7.1 install with nothing else added."

this is not in contradiction with content of not-reportable:
"A kernel problem occurred, but your kernel has been tainted (flags:GW). Kernel maintainers are unable to diagnose tainted reports."

That kernel is tainted, doesn't imply something had to be loaded into it, however I agree that this message can be confusing in this situation.
'''

I thought about closing it as NOTABUG, but I think it may be fine to explain user the situation. In fact abrt does nothing wrong in my point of view, but less informed user may explain this as in case of bug 1452160.

juliusmilan:
if someone will not like this solution I will explain this is solution for bug opening comment and ask him to open a new bug, maybe about enabling of reporting of W flagged oopses, but with exactly written demands and with respect to my arguments.

jfilak:
@juliusmilan  Do you still want to merge this PR?

juliusmilan:
@jfilak yes, I think it is good PR anyway, even when it doesn't solve problem with reporting tainted flags, it is good for user experience. I will let the bug open with needinfo and will close just its Fedora clone yet.

xsuchy:
retest this

mkutlak:
retest this please

PR #1246, opened 2017-05-19
Abrt action save container data fix

matejkudera:
Changing  abrt-action-save-container-data location from bin to libexec.

### DIFF
--- a/src/daemon/Makefile.am
+++ b/src/daemon/Makefile.am
@@ -7,8 +7,7 @@ bin_SCRIPTS = \
     abrt-handle-upload
 
 bin_PROGRAMS = \
-    abrt-action-save-package-data \
-    abrt-action-save-container-data
+    abrt-action-save-package-data
 
 sbin_PROGRAMS = \
     abrtd \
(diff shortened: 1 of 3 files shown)


trapas:

jfilak:
If you change fs location of abrt-action-save-container-data, you should also change fs location of abrt-action-save-package-data.

By the way, proper segregation of changes would be:
- spec file changes
- all other changes

What is the reason for this PR? (don't reply, update Commit messages, please).

xsuchy:
@jfilak I do not get why you wanna split it into more commits? I thought you are in favour of merged commits. And in my opinion this should be done in one commit, because one change without the other will not work. So it should be atomic.

jfilak:
> I thought you are in favour of merged commits.

Your thoughts are wrongs. I am in favor of meaningful commits.

> And in my opinion this should be done in one commit, because one change without the other will not work. So it should be atomic.

True but we were always separating spec file changes from the rest of changes because it is easier to use such commits in downstream repositories (and cherry pick them to RHEL6/7).

I usually go through the history of a project and take a look at old commits before I create new ones.

src/daemon/Makefile.am
```diff
@@ -7,16 +7,18 @@ bin_SCRIPTS = \
     abrt-handle-upload
 
 bin_PROGRAMS = \
-    abrt-action-save-package-data \
```
jfilak: You have removed `abrt-action-save-package-data` from `bin_PROGRAMS` but you haven't added it to `libexec_PROGRAMS`

PR #1252, opened 2017-06-15
a-d-journal-core: use pid of crashed process in dumpdir name

mhabrnal:
Without this commit pid of abrt-dump-journal-core is used in a name
of created problem directory instead of pid of crashed process.

Requires https://github.com/abrt/libreport/pull/495

### DIFF
--- a/src/plugins/abrt-dump-journal-core.c
+++ b/src/plugins/abrt-dump-journal-core.c
@@ -71,6 +71,7 @@ struct crash_info
     char *ci_executable_path;          ///< /full/path/to/executable
     const char *ci_executable_name;    ///< executable
     uid_t ci_uid;
+    pid_t ci_pid;
 
     struct field_mapping *ci_mapping;
     size_t ci_mapping_items;
@@ -219,6 +220,13 @@ abrt_journal_core_retrieve_information(abrt_journal_t *journal, struct crash_inf
         return -EINVAL;
     }
 
+    /* This is not fatal, the pid is used only in dumpdir name */
+    if (abrt_journal_get_int_field(journal, "COREDUMP_PID", &(info->ci_pid)))
+    {
+        log_notice("Failed to get PID from journal message.");
+        info->ci_pid = getpid();
+    }
+
     char *proc_status = abrt_journal_get_string_field(journal, "COREDUMP_PROC_STATUS", NULL);
     if (proc_status == NULL)
     {
@@ -322,7 +330,7 @@ save_systemd_coredump_in_dump_directory(struct dump_dir *dd, struct crash_info *
 static int
 abrt_journal_core_to_abrt_problem(struct crash_info *info, const char *dump_location)
 {
-    struct dump_dir *dd = create_dump_dir(dump_location, "ccpp", /*fs owner*/0,
+    struct dump_dir *dd = create_dump_dir_ext(dump_location, "ccpp", info->ci_pid, /*fs owner*/0,
             (save_data_call_back)save_systemd_coredump_in_dump_directory, info);
 
     if (dd != NULL)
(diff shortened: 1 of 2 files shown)


src/plugins/abrt-dump-journal-core.c
```diff
@@ -219,6 +220,13 @@ abrt_journal_core_retrieve_information(abrt_journal_t *journal, struct crash_inf
         return -EINVAL;
     }
 
+    /* This is not fatal, the pid is used only in dumpdir name */
+    if (abrt_journal_get_int_field(journal, "COREDUMP_PID", &(info->ci_pid)))
+    {
+        log_notice("Failed to get PID from journal message.");
+        info->ci_pid = getpid();
```
jfilak: Do we really want to lie? Would not it be better to just use no PID or max PID?

jfilak:
Good catch! The restart feature of abrt-applet does not work - hence I consider it a high priority issue. Can you please add a test case for the issue?

jfilak:
But we already save PID of the crashed process:
https://github.com/mhabrnal/abrt/blob/2aa79cd6a3fb7c0d2d4fbc26c1b2337ab6019668/src/plugins/abrt-dump-journal-core.c#L47

jfilak:
D'oh, this pull request is only about dump dir name. What was the problem with it?

src/plugins/abrt-dump-journal-core.c
jfilak: Ignore this comment, please.

marusak:
@jfilak with abrt-journal-core instead of abrt-ccpp the dumdir names look like this:
`ls /var/spool/abrt

 ccpp-2017-06-15-14:57:31.525526-10777
  ccpp-2017-06-15-15:00:50.25927-10777 
  last-ccpp
ccpp-2017-06-15-15:00:37.775260-10777 
 ccpp-2017-06-15-15:10:57.775036-10777`

I think you can see the theme there.

`ps -p 10777

  PID TTY          TIME CMD
 10777 ?        00:00:00 abrt-dump-journ
`

Also we talked with @mhabrnal about the wierd dot thing. Not sure if he looked at it. It seems to be some part of time. Or do you know what it is?

However I think this need to be fixed on another place as well. See this image:
https://cloud.githubusercontent.com/assets/3228183/26103879/2684b81c-3a44-11e7-8001-cd9e592adc68.png
The journal message has wrong pid. @mhabrnal Does it solve this as well?

marusak:
> Also we talked with @mhabrnal about the wierd dot thing. Not sure if he looked at it. It seems to be some part of time. Or do you know what it is?

Ok, those are miliseconds

>However I think this need to be fixed on another place as well. See this image:
https://cloud.githubusercontent.com/assets/3228183/26103879/2684b81c-3a44-11e7-8001-cd9e592adc68.png
The journal message has wrong pid. @mhabrnal Does it solve this as well?

Well, this is not happening anymore. Wierd, but cannot reproduce.

mhabrnal:
retest this please

mhabrnal:
Test added.

mhabrnal:
retest this please

mhabrnal:
retest this please

jfilak:
Those miliseconds are necessary for avoiding race conditions.

That PID was used only for the dump directory name - you can treat it a random identifier. If you need to display PID of the crashed process, then get it from the relevant dump directory - the information is stored  in the file 'pid' - there is also the file global_pid that contains pid as seen from init's PID namespace.

So the question is why we need to fix the name? Who complained about that?

marusak:
> you can treat it a random identifier

Not really random, is it?:)

> So the question is why we need to fix the name? Who complained about that?

I did. If it is not necessary there, we can omit it. If we want to keep it there, at least we should put there something meaningful.

Do you have any problem with changing this?

jfilak:
> Not really random, is it?:)

True. The key point is - Do not even think about building a logic based on dump directory names. Dump directory names are subject to change and are generated by ad-hoc algorithms and every tool can create names for their dump directories their way.

> Do you have any problem with changing this?

No, I don't. I was just wondering why this PR exists.

PR #1253, opened 2017-06-20
Fix integration tests

mkutlak:


tests/runtests/abrt-python/watch_expected
```diff
@@ -1 +1 @@
-<problem.Ccpp (will_segfault killed by SIGSEGV)>
```
mhabrnal: looks like a bug in abrt-journal-core

mhabrnal: fixed by https://github.com/abrt/abrt/commit/cc9c66d11d96ac180a2e927a089b6fc75913a8a6

PR #1254, opened 2017-06-21
cli,dbus: Allow polkit to be optional at build time

wlindauer:
This might blow your mind, but my use case for ABRT is on an embedded platform. You might have figured out by now from my pattern of pull requests that I want to be able to remove as many optional features as possible.

### DIFF
--- a/src/cli/abrt-cli-core.c
+++ b/src/cli/abrt-cli-core.c
@@ -23,13 +23,17 @@
 /* It is not possible to include polkitagent.h without the following define.
  * Check out the included header file.
  */
+#ifdef HAVE_POLKIT
 #define POLKIT_AGENT_I_KNOW_API_IS_SUBJECT_TO_CHANGE
 #include <polkitagent/polkitagent.h>
+#endif
 
 int g_cli_authenticate;
 
+#ifdef HAVE_POLKIT
 static PolkitAgentListener *s_local_polkit_agent = NULL;
 static gpointer s_local_agent_handle = NULL;
+#endif
 
 /* Vector of problems: */
 /* problem_data_vector[i] = { "name" = { "content", CD_FLAG_foo_bits } } */
@@ -126,6 +130,7 @@ char *hash2dirname_if_necessary(const char *input)
 
 void initialize_polkit_agent(void)
 {
+#ifdef HAVE_POLKIT
     GError *error = NULL;
     PolkitSubject *subject = polkit_unix_process_new_for_owner(
                                 getpid(),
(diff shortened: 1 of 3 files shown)


trapas:

src/cli/abrt-cli-core.c
```diff
@@ -148,13 +153,18 @@ void initialize_polkit_agent(void)
     }
 
     g_object_unref(subject);
+#else
+    printf("Polkit support is currently disabled\n");
```
wlindauer: I am not attached to this string if you want it to say something different.

jfilak: Maybe we should rather use the common logging functions: https://github.com/abrt/libreport/blob/master/src/include/internal_libreport.h#L578

I propose to log this message as info message. IMHO it is not necessary to issue the message every time a user runs abrt-cli. Thoughts?

src/dbus/abrt-polkit.c
```diff
@@ -90,22 +94,31 @@ static PolkitResult do_check(PolkitSubject *subject, const char *action_id)
...
 
 PolkitResult polkit_check_authorization_dname(const char *dbus_name, const char *action_id)
 {
+#ifdef HAVE_POLKIT
     glib_init();
 
     PolkitSubject *subject = polkit_system_bus_name_new(dbus_name);
     return do_check(subject, action_id);
+#else
+    return PolkitYes;
```
jfilak: I would print out a warning message because this is harmful - everyone has access to private data of all other users.

src/dbus/abrt-polkit.c
```diff
@@ -90,22 +94,31 @@ static PolkitResult do_check(PolkitSubject *subject, const char *action_id)
...
+#ifdef HAVE_POLKIT
     glib_init();
 
     PolkitSubject *subject = polkit_unix_process_new_for_owner(pid,
             /*use start_time from /proc*/0,
             /*use uid from /proc*/ -1);
 
     return do_check(subject, action_id);
+#else
+    return PolkitYes;
```
jfilak: I would print out a warning message because this is harmful - everyone have access to private data of all other users.

src/cli/abrt-cli-core.c
wlindauer: I agree with using log levels. I just didn't know how you preferred it to be done and now I do. I will push a new patch shortly. Is being at INFO level good enough to not spam the user? Or do you prefer that it only prints out the first time abrt-cli is run?

src/dbus/abrt-polkit.c
wlindauer: Similar question. Print every run or only the first run?

wlindauer:
Were there any further changes that I need to make for this to be accepted?

jfilak:
@wlindauer I apologize for the delay. I was too busy with my day job. I going to merge this PR. Thank you!

wlindauer:
No worries. I understand. I just didn't want you to be waiting on anything from me. Thanks.

PR #1255, opened 2017-07-11
testsuite: fix time and config

mkutlak:


tests/runtests/dbus-configuration/runtest.sh
```diff
@@ -307,11 +309,19 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Int32"
+        # save config
+        cp -v $abrt_conf $abrt_conf".backup"
```
mhabrnal: please use rlFileBackup instead
http://file.brq.redhat.com/~szidek/beakerlib/#rlFileBackup

tests/runtests/dbus-configuration/runtest.sh
```diff
@@ -307,11 +309,19 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Int32"
+        # save config
+        rlFileBackup $abrt_conf
+        # change config (max size was reduced in aux/pre.sh due to limited available space on VM)
+        sed -i "s/MaxCrashReportsSize.*=.*\d*/MaxCrashReportsSize = 5000/" $abrt_conf
```
mhabrnal: Would be better to use augtool to do that. Something like this:
```
 augtool set /files/etc/abrt/abrt.conf/MaxCrashReportsSize 5000
```

mhabrnal:
thank you!

PR #1260, opened 2017-08-22
Containers with journal

marusak:
Now crashes from containers are not recognized when abrt-journal-core is running.
This PR fixes it.
It saves two now values:
1) cmdline, which later in postprocess executes abrt-action-save-container-data
2) mountinfo, which is necessary for abrt-action-save-container-data

Test would be lovely. abrt/tests/runtests/docker-essentials should be run with abrt-ccpp and abrt-journal-core as well.


marusak:
Looking at this I just could use map https://github.com/abrt/abrt/blob/master/src/plugins/abrt-dump-journal-core.c#L34 to add this items. Will rework it.

mkutlak:
retest this please

marusak:
Fixed, rebased and also test added.

Issue #891, opened 2014-12-16
Capture unpackaged crashers by default


hadess:
Unpackaged or unknown to Fedora. This would allow users to gather debugging information for system problems with third-party drivers or applications, and make it easier to discard data about those crashers without dropping to the command-line.

jfilak:
@hadess I'm not sure what you mean by the last sentence. I guess you want ABRT to delete a journald message that was used to create an abrt problem, right? So *gnome-abrt* should become an unofficial GUI for *coredumpctl*, right?

hadess:
I'd want to remove the journald entries if the abrt entry was removed as well. For example, being able to remove development crashers that I already handled locally.

But the main goal would be for unpackaged crashers to appear in the UI, so it was possible to analyse them and identify them, even for third-party applications and utilities.

jfilak:
OK. I had to ask because such a thing is not possible now. ABRT doesn't execute any hook when deleting a problem, it just remove the related problem directory.

hadess:
Being able to show it would be a first step, being able to delete it from the list (eg. from the view) would the 2nd one. I don't even know if one can remove core dumps from journald right now (we might not be able to, as it would modify the integrity of previous logs, break audit).

jfilak:
You can at least delete a coredump related to a particular journal message because systemd-coredump stores coredump files outside the journal.

mtoman:
Even though the change is already implemented I need to say this is a **bad** idea. I can not see a single benefit (the suggested gui-frontend-to-coredumpctl is absolutely out of scope of the project). On the other hand this will create tons of unreportable entries in ABRT GUI that are only going to confuse users.

mbrysa:
Automated Bug **Reporting** Tool - who are you going to report crashes of third-party apps to?
I'm for reverting this.

hadess:
If the goal is not to help developers and QE, then we can rip out half of gnome-abrt. We even made design decisions based on the fact that it wasn't only a bug reporting tool for end users.

PR #1266, opened 2017-09-15
Allow sending uReports for crashes from unpackged executables

mhabrnal:
Requires https://github.com/abrt/libreport/pull/503

### DIFF
--- a/src/plugins/abrt-action-ureport
+++ b/src/plugins/abrt-action-ureport
@@ -130,6 +130,7 @@ if __name__ == "__main__":
         report_type = dd.load_text("analyzer", 0)
 
     core_backtrace_exists = dd.exist("core_backtrace")
+    packaged = dd.exist("package")
     reported_to = dd.load_text("reported_to", DD_FAIL_QUIETLY_ENOENT)
     ureports_counter = try_parse_number(dd, "ureports_counter")
     count = try_parse_number(dd, "count")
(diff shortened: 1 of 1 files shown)


jfilak:
If a dump directory does not contain the file `package`, then the problem occurred in an unpackaged program, right?

src/daemon/abrt-action-save-package-data.c
```diff
@@ -283,8 +290,12 @@ static int SavePackageDescriptionToDebugDump(const char *dump_dir_name, const ch
 
         char *unpackaged_msg = xasprintf("Crash in unpackaged executable '%s'", executable);
         dd_save_text(dd, FILENAME_UNPACKAGED, unpackaged_msg);
-
         free(unpackaged_msg);
+
+        char *unpackaged_component = xasprintf("%s-unpackaged", basename);
+        dd_save_text(dd, FILENAME_COMPONENT, unpackaged_component);
```
jfilak: Please, do not create the file `component` with artificial names - that would lead to a confusion. Have you checked if there are EVENT conditions with `COMPONENT!=`?

src/daemon/abrt-action-save-package-data.c
```diff
@@ -274,6 +274,13 @@ static int SavePackageDescriptionToDebugDump(const char *dump_dir_name, const ch
         goto ret; /* return 1 (failure) */
     }
 
+    /* Check well-known interpreter names */
```
jfilak: Please, help me to understand how you check `well-known` interpreter names here.

src/daemon/abrt-action-save-package-data.c
xsuchy: We are doing this, because we want to have FAF, which accept reports from unpackaged code. In this case, in FAF, we want to create new component when such report arrives.
On FAF, we need to somehow distinguish between report for which we just do not have downloaded rpm file and between report which was never packaged and we need to create an artificial component in FAF. For that we decided to took basename of that $0.

jfilak: Alright then. So can you create the artificial component on FAF when it receives an unpackaged uReport? IMHO it would be a less error-prone approach because storing this information in a dump directory can confuse other ABRT tools.

marusak: We can do that easily in FAF. But I believe that we need to edit also satyr(? or other component) where uReport is created, since it requires component to be present.
Maybe do not create new element in dd, but when uReport is created and `component` element is missing, then create an artificial one?

juliusmilan:
I think, it is also needed to add the configuration option to turn on/off also creation of problem directories (not just reporting which is done in libreport's PR), since it could easily happen that user will experience flood of undesired ABRT reports. And IMO it should be off by default.
Next it would be great to be also able to blacklist (resp. whitelist) executable names when reporting of unpackaged executables is on (resp. off)

jfilak:
@juliusmilan  I believe you are calling for https://github.com/abrt/abrt/blob/master/src/daemon/abrt-action-save-package-data.conf#L21

and for https://github.com/abrt/abrt/blob/master/src/daemon/abrt-action-save-package-data.conf#L25

Capturing of unpackaged processes is on by default because of https://github.com/abrt/abrt/issues/891

PR #1269, opened 2017-10-12
cli-ng,hooks,python-problem: Allow python to be optional at build time

wlindauer:
This allows python2 and python3 to be independently used for builds that
only require using one or the other.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,12 +56,6 @@ IT_PROG_INTLTOOL([0.35.0])
 
 dnl ****** END ****************************************
 
-AM_PATH_PYTHON
-if test -z "$PYTHON"; then
-    echo "*** Essential program python not found" 1>&2
-    exit 1
-fi
-
 AC_PATH_PROG([ASCIIDOC], [asciidoc], [no])
 [if test "$ASCIIDOC" = "no"]
 [then]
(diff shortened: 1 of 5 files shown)


jfilak:
@mkutlak Do you know why there are no test results and no request for admins to look at this PR?

marusak:
@jfilak - because yesterday the server on which jenkins lives was down due to a physical moving of that server. We need to check if jenkins was properly started.

trapas:

mkutlak:
ok to test

configure.ac
```diff
@@ -80,13 +74,50 @@ AC_PATH_PROG([XMLTO], [xmlto], [no])
...
+        [echo "Then run configure again before attempting to build ABRT."]
+        [exit 1]
+    [fi]
+
+    PYTHON_CFLAGS=`python-config --cflags 2> /dev/null`
+    PYTHON_LIBS=`python-config --libs 2> /dev/null`
+
+    AC_SUBST([PYTHON2_PREFIX], ['${prefix}'])
+    AC_SUBST([PYTHON2_EXEC_PREFIX], ['${exec_prefix}'])
+
```
mhabrnal: These lines are placed twice in the document (line 209). Please remove the next ones.

mhabrnal:
Otherwise looks good to me.

configure.ac
wlindauer: Good catch. I updated the commit.

mhabrnal:
@wlindauer I tried to compile abrt without python2 and it doesn't work.

```
make[4]: *** No rule to make target 'abrt_exception_handler.py.in', needed by 'abrt_exception_handler.py'.  Stop.
```

You have to add some additional condition to `./src/hooks/Makefile.am` file.

wlindauer:
I am unsure what to do here. I do not encounter this error when building without python2 in my environment. Here are the configure options I pass in my environment. "--enable-native-unwinder --without-python2 --disable-static  --without-bodhi --without-polkit --without-pythondoc --without-pythontests --without-rpm --without-selinux --enable-nls" Can you give any other configuration to hit this failure?

mhabrnal:
Thank you, I've fixed it and pushed this commit https://github.com/abrt/abrt/pull/1302/commits/39ce2f36b8a1b813ea5ae371f7a2735c4dcacd15

PR #1271, opened 2017-10-20
testsuite: add tests for processing of unpackaged executables

juliusmilan:


tests/runtests/aux/lib.sh
```diff
@@ -121,6 +121,17 @@ function generate_python3_exception() {
     su -c will_python3_raise $1
 }
 
+function generate_crash_unpack() {
+    rlLog "Generate unpackaged crash"
```
jfilak: Maybe, change directory to a safe directory like /tmp.

juliusmilan: yes, you're right, good point, thank you

juliusmilan:
oh, I forgot to mention: "updated" :) few days ago

PR #1283, opened 2018-01-22
correctly parse buildrequires from spec file

xsuchy:


marusak:
Hmm, two things:
1) It outputs the list with quotation marks. So for example using `./autogen.sh sysdeps --install` does not work
2) It is missing some components. Such as `xz`, `tar`, `libreport-plugin-rhtsupport` and many more. Or does it only parse buildrequires?

jfilak:
I do understand that you do not like the one-liner I wrote some time ago, but the commit nor the PR does not say what you are trying to fix.

xsuchy:
@jfilak  There is a problem with that grep. Because it will print e.g., `BuildRequires: dbus-1-glib-devel` which should be printed only on suse. Yum will happily ignore packages which does not exists, but DNF fails (which is correct behaviour).

xsuchy:
@marusak Yes, I forgot --requires. 
And I will fix the --install  as well.

xsuchy:
updated

marusak:
Ok, it works now. It does not show suse (build)requires and also it provides version for package. @jfilak some comments?

jfilak:
@marusak It makes it more reliable. However, it requires the binary rpmbuild - which is not available by default. If it does not cause problems in the test environment, I am OK with this commit.

@xsuchy You are right but `--setopt=strict=0` should disable that behaviour - if dnf does not ignore non-existing packages, then I call dnf broken:
https://github.com/xsuchy/abrt/commit/64faee2606f2722923ee319ad7689a9b43a18f51#diff-f88e95eed7716507cfe3a54a2c9dd382

autogen.sh
```diff
@@ -42,10 +40,9 @@ case "$1" in
         ;;
     "sysdeps")
             DEPS_LIST=$(list_build_dependencies)
-
             if [ "$2" == "--install" ]; then
                 set -x verbose
-                sudo dnf install --setopt=strict=0 $DEPS_LIST
+                eval sudo dnf install --setopt=strict=0 $DEPS_LIST
```
jfilak: Hmm, why eval?

jfilak:
Ehm, it is not rpmbuild but rpmspec which is part of the package rpm even on suse. So, there should not be any problem with the patch.

jfilak:
@xsuchy I would not mind if you were a bit more verbose in commit messages :)

autogen.sh
xsuchy: That's tricky :)  When DEPS_LIST="foo 'bar >= 1.0'" then `sudo dnf install $DEPS_LIST` expands to `sudo dnf install foo 'bar' '>=' '1.0'` but if you eval it first (before calling sudo) you will get `sudo dnf install foo 'bar >= 1.0'` and that is what we want.

mkutlak:
test this please

mkutlak:
test this please

Issue #1270, opened 2017-10-19
Crash from different containers should not be removed as duplicate


marusak:
When some crash happens in container X and then the same crash happens in container Y, should only one dumdir be created or two? It can be confusing, if somebody wants to know, what crashed in a specific container.
```
# ls /var/spool/abrt 
> <nothing>
$ docker run fedora /usr/bin/bash -c "timeout -s ABRT 1 sleep 10"
> 62020d004ede102bb686266c720f7967f2e2064473e35b703e329b4a14310f6b
# ls /var/spool/abrt 
> ccpp-2017-10-19-12:40:41.11277-16927
$ docker run fedora /usr/bin/bash -c "timeout -s ABRT 1 sleep 10"
> 406e8b3801bf996f0a307831a5717f1109a64b31c76ff82919c0df929318573a
# ls /var/spool/abrt 
> ccpp-2017-10-19-12:40:41.11277-16927

# cat ccpp-2017-10-19-12:40:41.11277-16927/container_id
> 62020d004ede

jfilak:
Good point! I have been thinking about improvements in how ABRT handles duplicates for some time now.

I don't think it is useful to keep both coredump files and other debugging files, but ABRT should not remove the dumpdir entirely - some files must be preserved and a link between the first occurrence and the other occurrence must be established. gnome-abrt / abrt-cockpit must display only the first occurrences, though, the history could be available.

The complete removal of dump directories cause problems to abrt-applet which can restart a crashed application - only it does not use the same command line to re-start the process as it uses the first captured command line - what a surprise for users!

marusak:
Can we put some thoughts together how handling duplicates could work?

> I don't think it is useful to keep both coredump files and other debugging files, but ABRT should not remove the dumpdir entirely - some files must be preserved and a link between the first occurrence and the other occurrence must be established. 

Agreed.  Should it be done as simlinks? (simple to implement and only symlink files, that are the same) Or some more elaborated way, such as new element 'duplicate_of' and path to the original dumpdir? But in both cases there is problem if original DD gets deleted. Or any other ideas?

> gnome-abrt / abrt-cockpit must display only the first occurrences, though, the history could be available.

abrt-cockpit uses dbus API and gnome-abrt as well(?). So we only need to teach ABRT's dbus API to handle this duplicates and return problems adequately.

xsuchy:
We can make this configurable.

PR #1284, opened 2018-01-24
Duplicate container crashes

marusak:
Related to #1270 
This is the simplest solution. I do **not** insist on merging this PR. It is more like  _conversation starter_, so we can find the best solution.

This keeps the whole dumpdir event when the same crash happened in two different containers. But it deduplicates the same problem from one container.

There are many options, how to improve this, but in each and every of theme  is some catch, that is way to complex to solve.

Options @jfilak @mhabrnal @xsuchy ?

### DIFF
--- a/src/daemon/abrt-handle-event.c
+++ b/src/daemon/abrt-handle-event.c
@@ -229,6 +229,7 @@ static int is_crash_a_dup(const char *dump_dir_name, void *param)
     type = dd_load_text(dd, FILENAME_TYPE);
     free(executable);
     executable = dd_load_text_ext(dd, FILENAME_EXECUTABLE, DD_FAIL_QUIETLY_ENOENT);
+    char *container_id = dd_load_text_ext(dd, FILENAME_CONTAINER_ID, DD_FAIL_QUIETLY_ENOENT);
     dup_uuid_init(dd);
     dup_corebt_init(dd);
     dd_close(dd);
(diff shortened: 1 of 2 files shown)


mhabrnal:
@marusak I have the same opinion about this issue as the previous week. I think this solution is easiest and it's a good enought for the beginning.

jfilak:
OK. I we can merge this PR once you describe the use case when you want to see the crash appearing in different containers. Please do not think it is a common knowledge. I had to think about it at least for a minute and I am not sure I found the same reason for this PR as you.

marusak:
Ok, I see what you mean, that usually users/admins do not care about specific containers that much. However this makes more sense when I assume user using Cockpit.
Story time:
I, as a devops, test my App in containers. Testing is sometimes interupted and does not finish. It happened again so I use Cockpit to quickly check about the container.
I see this:
ok
However, If I looked further I would see that in previous container there was problem caught by ABRT.
wrong (of course, there could be like 400 of these containers, not only 3 :) )
I would immediately know, that there is something happening if I saw the exclamation mark on containers, in which problem occurred.

jfilak:
Good, we are on the same page. Please, try to update the commit message with this particular use case and we can merge this PR.

marusak:
@jfilak updated

mkutlak:
test this please

PR #1285, opened 2018-01-30
Let containers know about their coredumps

jfilak:
The commits of this pull request should allow image/container developers use the tool abrt-dump-journal-core to get information about crashes in their containers.

Actually, this pull request just adds the possibility to read core dump files from all systemd journals. It does that because when you run systemd in a container you have two journals:
- the host's journal
- the container's journal

Without this PR, the tool abrt-dump-journal-core reads only the latter one - the container's journal.

After merging this journal, you should be able to use the parametre "-a" to read also the host's journal where coredump files end up.

The next steps is to introduce another option to filter out all coredump files that are not from the container's hostname:
https://github.com/systemd/systemd/commit/f45b8015513d38ee5f7cc361db9c5b88c9aae704#diff-2a71e915f668d9cf67c1ebd2c2324d23

### DIFF
--- a/src/plugins/abrt-dump-journal-core.c
+++ b/src/plugins/abrt-dump-journal-core.c
@@ -17,6 +17,10 @@
 
 #define ABRT_JOURNAL_WATCH_STATE_FILE VAR_STATE"/abrt-dump-journal-core.state"
 
+enum {
+    ABRT_CORE_PRINT_STDOUT = 1 << 0,
+};
+
 /*
  * A journal message is a set of key value pairs in the following format:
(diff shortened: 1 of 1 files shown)


mkutlak:
test this please

jfilak:
Should I do something to make this PR merged into master?

xsuchy:
@mhabrnal will do the review as soon as possible

mhabrnal:
@jfilak could you, please, be more verbose in the pull request's description? I am not sure what the pull request does/fixes.

mhabrnal:
Look good to me. Thank you!

PR #1296, opened 2018-03-16
Add a new element 'interpreter' for python problems 

### LINKED ISSUE #1294: Show Python version with "type: Python3" bugzilla reports
When there is an ABRT bugzilla report with

    type:           Python3

Often, Fedora package maintainers reassign to the python3 component. However, I struggle to find out what version-release the python3 package has. Can that be shown? For us, it's more important than the kernel version (that one is shown).

See for example https://bugzilla.redhat.com/show_bug.cgi?id=1551420

It would also be great to see it in related packages in https://retrace.fedoraproject.org/faf/reports/1850656/

The same probably applies to python2 problems.

mhabrnal:
The element 'interpreter' contains rpm package
of an interpreter which ran a script which raised
an exception.

Fixes #1294

### DIFF
--- a/src/plugins/python3_event.conf
+++ b/src/plugins/python3_event.conf
@@ -8,6 +8,10 @@ EVENT=post-create type=Python3 remote!=1
             exit 1
         fi
         abrt-action-analyze-python
+        # save Python3 package version
+        for line in $(rpm -qf $(which $(cut -d' ' -f1 < cmdline)) 2>/dev/null); do
+             echo -n $line > interpreter
+        done
 
 EVENT=report_Bugzilla type=Python3 component!=anaconda
         test -f component || abrt-action-save-package-data
--- a/src/plugins/python_event.conf
+++ b/src/plugins/python_event.conf
@@ -8,6 +8,10 @@ EVENT=post-create type=Python remote!=1
             exit 1
         fi
         abrt-action-analyze-python
+        # save Python2 package version
+        for line in $(rpm -qf $(which $(cut -d' ' -f1 < cmdline)) 2>/dev/null); do
+             echo -n $line > interpreter
+        done
 
 EVENT=report_Bugzilla type=Python component!=anaconda
         test -f component || abrt-action-save-package-data
--- a/tests/runtests/dumpdir_completedness/runtest.sh
+++ b/tests/runtests/dumpdir_completedness/runtest.sh
@@ -35,7 +35,7 @@ PACKAGE="abrt"
 DDFILES="abrt_version analyzer architecture cmdline component count cpuinfo executable hostname kernel last_occurrence os_release package pkg_arch pkg_epoch pkg_name pkg_release pkg_version pkg_vendor pkg_fingerprint reason time type uid username uuid os_info runlevel"
 
 CCPP_FILES="core_backtrace coredump dso_list environ limits maps open_fds pid pwd cgroup global_pid  proc_pid_status"
-PYTHON_FILES="backtrace"
+PYTHON_FILES="backtrace interpreter"
 
 rlJournalStart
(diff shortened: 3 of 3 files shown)


jfilak:
You don't need to parse cmdline twice: https://github.com/abrt/abrt/blob/master/src/daemon/abrt-action-save-package-data.c#L307

jfilak:
Are you going to update also the libreport Bugzilla templates?

mhabrnal:
> Are you going to update also the libreport Bugzilla templates?

If 'the libreport Bugzilla templates' is a formatting template for bugzilla's description (bugzilla_format.conf), I don't plan to update it. Can you please tell me why I should do that?

> You don't need to parse cmdline twice: https://github.com/abrt/abrt/blob/master/src/daemon/abrt-action-save-package-data.c#L307

I know about this part of a code. If I understand it right, the if branch is executed only if an interpreter is in a basename. So, if the basename is extracted right (interpreter is only in cmdline), the if branch is not executed at all. That's the reason why I added creating of a interpreter element into event conf file.

jfilak:
>  Can you please tell me why I should do that?

Because @hroncok wants to see the version of Python in Bugzilla reports and I think it would be cool provide him with the information somewhere hi can easily spot it.

> I know about this part of a code. If I understand it right, the if branch is executed only if an interpreter is in a basename. So, if the basename is extracted right (interpreter is only in cmdline), the if branch is not executed at all. That's the reason why I added creating of a interpreter element into event conf file.

No one prevents you from refactoring abrt-action-save-package-data. IMHO better to refactor it then introduce the same logic again in a different tool - more over in a place where it is not expected to be -  event files are for hacking :)

mhabrnal:
>> Can you please tell me why I should do that?

> Because @hroncok wants to see the version of Python in Bugzilla reports and I think it would be cool  provide him with the information somewhere hi can easily spot it.

Python version is provided in bugzilla's description without a need to change bugzilla's template

```
[Mon May 14 12:58:45 @root /var/spool/abrt/Python3-2018-05-14-12:58:42-10647]
(0) $ reporter-bugzilla -D .
Login is not provided by configuration. Please enter your https://bugzilla.redhat.com login: sadf
Password is not provided by configuration. Please enter the password for 'sadf': 
summary: [abrt] will-crash: module(): will_python3_raise:3:<module>:ZeroDivisionError: division by zero

Version-Release number of selected component:
will-crash-0.10-7.fc28

Additional info:
reporter:       libreport-2.9.5
cmdline:        /usr/bin/python3 /bin/will_python3_raise
crash_function: module
exception_type: ZeroDivisionError
executable:     /bin/will_python3_raise
interpreter:    python3-3.6.5-1.fc28.x86_64
kernel:         4.16.5-300.fc28.x86_64
runlevel:       N 5
type:           Python3
uid:            0

Truncated backtrace:
will_python3_raise:3:<module>:ZeroDivisionError: division by zero

Traceback (most recent call last):
  File "/bin/will_python3_raise", line 3, in <module>
    0/0
ZeroDivisionError: division by zero

Local variables in innermost frame:
__name__: '__main__'
__doc__: None
__package__: None
__loader__: <_frozen_importlib_external.SourceFileLoader object at 0x7f733e4fab00>
__spec__: None
__annotations__: {}
__builtins__: <module 'builtins' (built-in)>
__file__: '/bin/will_python3_raise'
__cached__: None

attachments:
 backtrace
 cgroup
 cpuinfo
 environ
 machineid
 mountinfo
 namespaces
 open_fds
```

hroncok:
If I see this in BZ:

```
Additional info:
...
interpreter:    python3-3.6.5-1.fc28.x86_64
...
```

Then I'm happy.

PR #1301, opened 2018-04-23
abrt-dump-oops: return last oops instead of first

dandedrick:
When picking an oops out of multiple occurrence picking the first one is
often wrong. Specifically for vmcore events the last oops is the most
interesting one so it makes the most sense to pick that one.

This choice may not always be correct but it seems like it will be
correct more often than when the first oops encountered is chosen.


trapas:

PR #1307, opened 2018-05-30
abrtd.service: force abrt-dbus to load changes from conf

mhabrnal:
If abrt-dbus is running and some changes are done in abrt.conf the
changes are not read and used till abrt-dbus timeouts. Timeout is set
to 133 seconds in 'dbus/org.freedesktop.problems.service'.
    
Killing abrt-dbus before abrtd is started forces abrt-dbus to read all
new configurations.

### DIFF
--- a/init-scripts/abrtd.service
+++ b/init-scripts/abrtd.service
@@ -8,6 +8,7 @@ After=livesys.service
 [Service]
 Type=dbus
 BusName=org.freedesktop.problems.daemon
+ExecStartPre=/usr/bin/bash -c "pkill abrt-dbus || :"
 ExecStart=/usr/sbin/abrtd -d -s
 
 [Install]
--- a/tests/runtests/abrtd-directories/runtest.sh
+++ b/tests/runtests/abrtd-directories/runtest.sh
@@ -45,7 +45,6 @@ rlJournalStart
         TmpDir=$(mktemp -d)
         pushd $TmpDir
     rlPhaseEnd
-
     rlPhaseStartTest
         # abrtd creates the dump location
         rlRun "systemctl restart abrtd"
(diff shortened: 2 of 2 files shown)


jfilak:
I guess this break abrt for cockpit which is based on abrt-dbus running all the time and keeping DBus connections open. @marusak is your cockpit integration listening for abrt-dbus signals?

marusak:
> is your cockpit integration listening for abrt-dbus signals?

Only for "docker crashes". For standard problems, journald is parsed (where ABRT writes), when in journal is recognized that it comes from ABRT, then we read more info from ABRT (through dbus).

For "docker crashes" we listen for `Crash` signal. https://github.com/cockpit-project/cockpit/blob/master/pkg/docker/containers-view.jsx#L246

jfilak:
@mhabrnal Can you just send the signal `SIGHUP` to `abrt-dbus` to reload the configuration?

mhabrnal:
@jfilak looks like `SIGHUP` has the same effect as `SIGTERM`. After `abrt-dbus` is killed signal `Crash` is sent anyway.

I did it the same way you did in ABRT's spec file scriplet https://github.com/abrt/abrt/blob/master/abrt.spec.in#L900

jfilak:
> @jfilak looks like SIGHUP has the same effect as SIGTERM. After abrt-dbus is killed signal Crash is sent anyway.

You are right, absolutely and definitely, signals without handler causes processes to terminate. That means that you need to add your handler and reload the configuration there - have a look into other abrt daemons for implementation. Once upon a time, you have a chance to code a little bit!

> I did it the same way you did in ABRT's spec file scriplet 

Sure, after an upgrade, the binary might not be the same, linked libraries can be different, communication protocols with other ABRT processes might be changed. You must restart the processes otherwise you start experiencing weird crashes and strange errors that cannot be reproduced.

But you update your system occasionally and ABRT gets updated even less often - that's completely different situation than that you want to deal with in the PR, right?

jfilak:
BTW, I installed my Fedora 27 more than 6 month ago so far I have detected 2 crashes - well done Fedora maintainers - (I use KDE :)!

mhabrnal:
@jfilak, dbus is not my strong area but if I understand it right, an initial state of abrt-dbus is not running. If it runs, it timeouts after 133 seconds of nonactivity and then it's turned to the initial state again. So, abrt-dbus is most of the time in the initial state.
What is so wrong with a solution which forces abrt-dbus to switch to the initial state before it timeouts (just in time a service abrtd is restarted which not happening very often)? Is it the main program loop is not ended properly or anything else which I am not able to get right now?

jfilak:
abrt-dbus exits after 133 only if there are no clients connected. @marusak 's cockpit plugin is a client. That client registers a DBus signal handler. If you kill the abrt-dbus process, the handler will stop working - however, I guess, no one will notice it - so it will look like there are no new crashes.

By, the way, abrtd re-loads configuration by calling `load_abrt_conf` in every event handler.

However, I did not realize that you restart abrt-dbus only upon restart of abrtd. My bad. I forgot that `abrtd` is running all the time - I was under the impression it is started upon every crash. Sorry.

PR #1308, opened 2018-06-13
ccpp: add %h and %e parameter into abrt-hook-ccpp

mhabrnal:
Without this commit core_pattern's parameters %h and %e was not
    translated at all.
    
    Due to 51531e9ea72af09571688a20f00e55cad5fe9c3a, %e has to be placed
    at the very end of the command in core_pattern.
    
    Example:
    If 'core_pattern = core.%h.%e.%p.%t' the result was
    core.%h.sleep.26284.1469805542 not core.myshostmane.sleep.26284.1469805542.
    
    Related to #1587891

### DIFF
--- a/src/hooks/abrt-hook-ccpp.c
+++ b/src/hooks/abrt-hook-ccpp.c
@@ -65,13 +65,13 @@ static struct dump_dir *dd;
  * %t - UNIX time of dump
  * %P - global pid
  * %I - crash thread tid
- * %e - executable filename (can contain white spaces)
+ * %h - hostname
+ * %e - executable filename (can contain white spaces, must be placed at the end)
  * %% - output one "%"
  */
 /* Hook must be installed with exactly the same sequence of %c specifiers.
- * Last one, %h, may be omitted (we can find it out).
  */
-static const char percent_specifiers[] = "%scpugtePi";
+static const char percent_specifiers[] = "%scpugtPIhe";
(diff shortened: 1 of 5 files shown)


jfilak:
Nice test. Good job! What about to add a test with `%e` substituted by a string with a space - `pthread_setname_np`?

jfilak:
Otherwise, I do find this PR ready to be merged ;)

mhabrnal:
@jfilak I tried `pthread_setname_np` to set name containing space and it didn't work well. So I decided to obtain executable filename from file followed by a symlink /proc/PID/exe (if it exists).

jfilak:
@mhabrnal are you going to update the test too?

jfilak:
@mhabrnal Actually, what you mean by "it didn't work well"? Are you going to write a note about this decision into the commit message or as comment in the code?

mhabrnal:
@jfilak If I set thread name containing spaces, an only first word is used. I am going to update test and commit message as well.

jfilak:
Put the %e parameter into quotes then.

jfilak:
Or you need to join all the argv values after the index 9 by space.

mhabrnal:
I tied the quotes and it did work as expected, the idea of joining argv values after the index 9 come to my mind to, but I decided to do it the same way as is done in systemd-coredump. I like this approach the most.

PR #1309, opened 2018-06-25
harvest_vmcore: Fix missing argument error during delete_and_close()

jdobina:
delete_and_close() requires a directory name argument and it is being called
without one. This argument is really not necessary though since the directory
name is already saved in the directory object (can be queried via the directory
object's name attribute), and it is the saved directory that is always deleted
regardless of the argument passed in.

Signed-off-by: Jake Daryll Obina <jake.obina@gmail.com>


trapas:

mhabrnal:
test this please

PR #1314, opened 2018-07-24
plugins: Replace vfork with posix_spawn

### LINKED ISSUE #1311: Do not use vfork
Call to function 'vfork' is insecure as it can lead to denial of service situations in the parent process. Replace calls to vfork with calls to the safer 'posix_spawn' function

mkutlak:
vfork is insecure as it can lead to denial of service situations in the parent process[1].

[1] https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152373

resolves #1311

### DIFF
--- a/src/plugins/abrt-retrace-client.c
+++ b/src/plugins/abrt-retrace-client.c
@@ -16,11 +16,14 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 #include "https-utils.h"
+#include <spawn.h>
 
 #define MAX_FORMATS 16
 #define MAX_RELEASES 32
 #define MAX_DOTS_PER_LINE 80
 #define MIN_EXPLOITABLE_RATING 4
+
+extern char **environ;
 
 enum
 {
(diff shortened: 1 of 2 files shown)


jfilak:
Wow, any particular reason for not using just fork?

mkutlak:
It is suggested to replace `vfork` with `posix_spawn`. There are some performance/efficiency benefits over `fork`.

https://github.com/rtomayko/posix-spawn#benchmarks

jfilak:
Execuse me, but what do you mean by "it"?  The page you linked in the commit message suggests to replace `vfork` with `fork`.

I would rather ask @dvlasenk about his opinion on this patch.

jfilak:
@xsuchy Can you go upstairs and ask Denis for guidance here, please?

mkutlak:
Here is feedback from @dvlasenk:

> Ok. I'm not familiar with posix_spawn(). If I need to guess how it works, it's just a vfork/exec under the hood or fork/exec.

> Basically, this change switches to using an additional API layer because it  does not trust that bare vfork would work correctly.

> IOW: the change is probably "correct" in a sense that it does not add bugs, but I would not be doing that change. I prefer less layers, not more. I trust vfork() works correctly. If I find a bug in it, I would report it to LKML and get it fixed.

> If I wouldn't care about performance and code for a very paranoid application, I would just eat performance penalty and use fork() instead of vfork(). abrt well may be in that situation - it does not fork 1000s of processes a second.

xsuchy:
As everybody has mixed feelings and I did not hear strong objection about this PR, but there are some objection against vfork(), I am going to merge this PR as this is probably the best we have.

jfilak:
Tehere were 2 against (me and Denys) and 1 for. IMHO The best solution was to use `fork()`. Almost no code changes versus introducing completely new thing that is unknow even to hardcore folks like Denys.

jfilak:
I still do not understand where @mkutlak found this function.

PR #1321, opened 2018-09-21
daemon: Fix double closed fd race condition

### LINKED ISSUE #1280: abrtd polls on invalid file descriptor (probably double close)
I ran into a case of abrtd polling on an invalid file descriptor:

```
poll([{fd=3, events=POLLIN|POLLPRI}, {fd=5, events=POLLIN}, {fd=6, events=POLLIN|POLLPRI}, {fd=9, events=POLLIN|POLLPRI}, {fd=14, events=POLLIN}], 5, -1) = 1 ([{fd=14, revents=POLLNVAL}])
```

It looks like FD 14 in question is a socket opened to a client:

```
(gdb) print *(GPollFD*)source->poll_fds->data
$18 = {fd = 14, events = 17, revents = 32}
(gdb) print *(GSourceCallback*)source->callback_data
$19 = {ref_count = 1, func = 0x562208686710 <abrt_server_output_cb>, data = 0x562208a7a320, notify = 0x0}
```

It seems like the issue is a double close somewhere. I found two double closes, I think, but I am not sure why these would actually trigger the error (i.e. the two close() calls should be right after each other). So there may be another double close somewhere.

mkutlak:
When a communication channel is set up between abrtd and abrt-server it uses
abrt_gio_channel_unix_new(). In that function there is a call g_io_channel_set_close_on_unref() [1].
This function sets whether to close a file/socket/whatever associated with the channel when channel recieves a final unref and is to be destroyed.

Calling a close() on fd associated with the channel before/after g_io_channel_unref()
created a double close() race condition when ABRT was processing a lot of crashes at the same time.

Thank you @benzea for the patch.

Fixes #1280
Related BZ#1416310

1 - https://developer.gnome.org/glib/stable/glib-IO-Channels.html#g-io-channel-get-close-on-unref

### DIFF
--- a/src/daemon/abrt-server.c
+++ b/src/daemon/abrt-server.c
@@ -364,7 +364,6 @@ static int run_post_create(const char *dirname, struct response *resp)
     g_main_loop_unref(context.main_loop);
     g_io_channel_unref(channel_signal);
     close(g_signal_pipe[1]);
-    close(g_signal_pipe[0]);
 
     log_notice("Waiting finished");
 
--- a/src/daemon/abrtd.c
+++ b/src/daemon/abrtd.c
@@ -114,7 +114,6 @@ static void stop_abrt_server(struct abrt_server_proc *proc)
 
 static void dispose_abrt_server(struct abrt_server_proc *proc)
 {
-    close(proc->fdout);
     free(proc->dirname);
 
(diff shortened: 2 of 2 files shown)


mkutlak:
@trapas retest this please

PR #1324, opened 2018-10-14
fix for MCE events: Bug 1613182 - abrt-cli ignores rsyslog host info …

clime:
…and incorrectly assumes that the receiving host had a problem

Also adds fixes for tests because host name in test input files now matters. Test input files can now have an optional suffix ".template" and contain tags `:HOSTNAME:` or `:FULL_HOSTNAME:` which gets replaced by actual values on a machine where tests are run by new `prepare-data` script.

Note that this addition newly requires a short or long hostname to be present before "kernel: " string. If it is not present, the message is ignored.


mkutlak:
test this please

xsuchy:
test this please

clime:
Should be ok now.

mkutlak:
LGTM. Great job!


Just two things that need fixing.

 - We usually try to follow the following format of GIT commits.
```
topic: Short description of changes made in the commit

Longer description of changes made in the commit*

Related: List of related Bugzillas*
```
This makes it easier for us to find old relevant commits in the git log history, and to create new releases of abrt.


 - And the last thing is coding style - `{}` brackets  after `if, else, ...` should be always placed on the new line.

clime:
Hello, I have added the commit with fixed formatting and the message. Can you check? I guess I should squash the commits in the end..

clime:
Hello, I have added the commit with fixed formatting and the message. Can you check? I guess I should squash the commits in the end..

mkutlak:
Thank you @clime

PR #1328, opened 2018-11-09
Port retrace client to libsoup

ernestask:
On the heavier side, but this enables using proxies for connections via glib-networking, which supports GNOME directly and other configuration means via libproxy, including the well-known environment variables (albeit a bit tricky to use, given that other modules take precedence, so even if you trick glib-networking into using libproxy on GNOME, you have to trick libproxy into not doing the same as well).


ernestask:
@trapas test this please

Issue #1085, opened 2015-11-10
abrt-applet tray icon missing


drorriss:
abrt-applet used to produce a tray icon (when a crash occurs) back in Fedora 20 and prior; when I upgraded to Fedora 22, the tray icon no longer appeared. This is a problem because I may not be at my computer when the notification bubble appears, so may be unaware that there was a crash when I return to my computer.

Is there a way to re-enable the tray icon? Was it removed? Or is it a bug that it is not appearing?

btw, I'm using the Mate Desktop environment.

jfilak:
The major desktop players have dropped XEmbed support, so the old school systray icon has stopped working. We need to port abrt-applet, the thing that shows the icon, to the new notification models.

ernestask:
> We need to port abrt-applet, the thing that shows the icon, to the new notification models.

I wonder what the new models are, since the applet has been using libnotify since 2009. `GNotification`? I have a working port (which also required porting to `GApplication`), but there’s not much of a difference, other than no longer using a less-than-maintained library, as the fdo spec support is still there, so it should play just as nicely with the myriad servers out there.

GTK+ no longer supports tray icons, meaning that notifications are indeed the way to go. Being away from keyboard should not be an issue if the shell/notification server of your choice does not disappear the notifications themselves.

PR #1329, opened 2018-11-13
applet: Port to GApplication/GNotification

ernestask:
This commit ports the applet from its current use of libnotify to
GNotification and, consequently, GApplication. That nets us persistent
notifications following the removal of the tray icon, since currently
they are lost when the applet goes down. Additionally, this makes the
applet D-Bus activatable for when the notifications are clicked, so that
the applet itself is respawned and gnome-abrt is launched. A bit
roundabout, but that’s that.

Closes https://github.com/abrt/abrt/issues/1085


ernestask:
@trapas test this please

ernestask:
@trapas shake it, baby

src/applet/abrt-applet-application.c
```diff
@@ -0,0 +1,1240 @@
...
+    {
+        return;
+    }
+
+    cachedir = g_get_user_cache_dir ();
+    dirlist_name = concat_path_file (cachedir, "abrt");
+
+    g_free (dirlist_name);
+
+    g_mkdir_with_parents (dirlist_name, 0777);
```
ernestask: Oh my god.

PR #1334, opened 2018-11-14
spec: turn on --enable-native-unwinder aarch64

mkutlak:
Turning on native-unwinder for aarch64 according to #1260074

Related to #1260074


xsuchy:
Nice.

PR #1335, opened 2018-11-26
testsuite: dbus-element-handling: Use external script

ernestask:
Invoking `python` inline can result in the Python 2 interpreter being
run, which might be problematic if the project was configured with
--without-python2, as the F29 test runner in particular does not have
python2-dbus installed, which is required for this test.


mkutlak:
Thank you.

All tests* should use Python3 only. If you find a test that still uses python/python2 feel free to port it to Python3.

*all tests on this list - https://github.com/abrt/abrt/blob/master/tests/runtests/aux/test_order

PR #1336, opened 2018-11-27
testsuite: reporter-upload-ssh-keys: Don’t test curl output

ernestask:
With the advent of the libssh backend in curl, some of the expected
output might not be there. There is no way to check if the specified key
file is being used just by reading the output.


ernestask:
Even checking failure messages doesn’t help:
https://github.com/curl/curl/blob/2e496108534ed20c82cf96f77443e81605f78f92/lib/ssh-libssh.c#L664
https://github.com/curl/curl/blob/2e496108534ed20c82cf96f77443e81605f78f92/lib/ssh-libssh.c#L2107

Other ideas welcome.

PR #1339, opened 2018-11-29
Port some test scripts to Python 3

ernestask:
See commits.


ernestask:
https://docs.python.org/2/library/basehttpserver.html for reference.

ernestask:
@trapas, shake it.

ernestask:
The pyftpdlib module was copied from https://github.com/giampaolo/pyftpdlib (minus the tests).

mkutlak:
> The pyftpdlib module was copied from [giampaolo/pyftpdlib](https://github.com/giampaolo/pyftpdlib) (minus the tests) (current master, maybe could’ve used a released version instead).

I think it would be better to just use the `python3-pyftpdlib` package for Fedora and then build that package for RHEL in copr.

ernestask:
> I think it would be better to just use the `python3-pyftpdlib` package for Fedora and then build that package for RHEL in copr.

Good idea. Dropped the in-tree copy.

mkutlak:
@trapas retest this please

tests/runtests/aux/config.sh
```diff
@@ -77,5 +77,6 @@ export PACKAGES="abrt \
                  libreport-plugin-logger \
                  libreport-plugin-mantisbt \
                  libreport-plugin-systemd-journal \
+                 python3-pyftpdlib \
```
mkutlak: We don't really need the `pyftpdlib` to be available for all tests, just for one.

Can you instead install the package in `rlPhaseStartSetup`? And maybe remove it in `rlPhaseStartCleanup`.

ernestask: I added a conditional removal, since blindly removing packages is not nice.

mkutlak:
LGTM. Thank you @ernestask

PR #1342, opened 2018-12-03
a-a-install-debuginfo: Clean cache if we need space

marusak:
When new debuginfo packages we going to be installed check if we have
space to store them. If we don't have enough space, but we have cached
enough packages, which are not now needed, make some space.

Fixes bz#811978
Requires: https://github.com/abrt/libreport/commit/ace233e8a1ca4036fcdaa1ba645de764b9a2806f

### DIFF
--- a/abrt.spec.in
+++ b/abrt.spec.in
@@ -50,7 +50,7 @@
     %define docdirversion -%{version}
 %endif
 
-%define libreport_ver 2.9.3
+%define libreport_ver 2.9.7
 %define satyr_ver 0.24
 
 Summary: Automatic bug detection and reporting tool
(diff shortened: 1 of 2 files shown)


xsuchy:
Why we do not use a simple LRU algorithm?

marusak:
> Why we do not use a simple LRU algorithm?

Algorithm for deleting is implemented in `abrt-action-trim-files` and I did not touch it. The only thing I do in this PR is looking how much space is occupied by packages, those are not needed in the current context.

xsuchy:
Test failing because of
```
Package gdb-headless-8.2-5.fc29.x86_64 is already installed.
No matching package to install: 'libreport-devel >= 2.9.7'
No matching package to install: 'libreport-gtk-devel >= 2.9.7'
No matching package to install: 'libreport-web-devel >= 2.9.7'
```
@marusak can you release and build new libreport, please?

marusak:
Yes, I am aware of this. I wanted to release new libreport today, but it seems that one test does not work on F29 and rawhide, so I am now debugging why it does not work.

mkutlak:
@trapas retest this please

PR #1352, opened 2019-01-10
Export workflow from reported_to over D-Bus

ernestask:
Needed if https://github.com/abrt/libreport/pull/549 is merged.


mkutlak:
You should also bump the `libreport_version` in a spec file. We will have to release new libreport after the abrt/libreport#549 is merged.

https://github.com/abrt/abrt/blob/66b100e1d0324e860d5f93077e5dc64cd901a0ad/abrt.spec.in#L53

ernestask:
I suppose the next version of libreport would be 2.10.0? I used that one.

mkutlak:
> I suppose the next version of libreport would be 2.10.0? I used that one.

Yeah, we can release new libreport as 2.10.0.

PR #1354, opened 2019-01-12
applet: application: Fix use after free

ernestask:
dirlist_name is erroneously freed before it’s used to create a
directory.


ernestask:
That was embarrassing.

PR #1361, opened 2019-01-29
applet: Add systemd service unit

ernestask:
Since using an autostart desktop file causes troubles on GNOME
(gnome-session chokes with D-Bus-activatable ones), having systemd
manage that instead probably makes more sense.


ernestask:
Still need to test this.

jfilak:
Don't forget about KDE :)

ernestask:
In what way do you mean? AFAIU, KDE runs things from /etc/xdg/autostart wholesale (I forgot already how GNOME does it, but it may look for a magic undocumented key), and this should work in a more DE-agnostic way.

In any case, msuchy brought KDE up as well, so I will make sure to test with whatever is at my disposal.

PR #1370, opened 2019-03-06
dbus: entry: Don’t claim truncation when there was none

ernestask:
When saving elements, if the file being copied exceeds size the limit,
the code will yell and bail. An issue with that is that the comparison
is not strict, which causes false positives when the actual file size is
equal to the limit.

### DIFF
--- a/src/dbus/abrt_problems2_entry.c
+++ b/src/dbus/abrt_problems2_entry.c
@@ -796,7 +796,7 @@ int abrt_p2_entry_save_elements_in_dump_dir(struct dump_dir *dd,
                 continue;
             }
 
-            if (r >= max_size)
+            if (r < item_stat.st_size && r >= max_size)
             {
                 error_msg("File descriptor was truncated due to size limit");
 


src/dbus/abrt_problems2_entry.c
```diff
@@ -796,7 +796,7 @@ int abrt_p2_entry_save_elements_in_dump_dir(struct dump_dir *dd,
                 continue;
             }
 
-            if (r >= max_size)
+            if (r < item_stat.st_size && r >= max_size)
```
xsuchy: Are you sure about that &&? What is in the value of r and st_size? I am trying to read the code and IIRC I would use ||

ernestask: ORing doesn’t really work here, as, in this particular case, `r == item_stat.st_size == max_size`, so the change would do nothing.

Basically, if there is no explicit limit for the dump directory (`limits->data_size == 0`), `max_size` becomes `item_stat.st_size`, and `r` is just the number of bytes read.

I think we can just do `r < item_stat.st_size`, since we really only need to know if truncation occurred due to some arbitrary limit, not if we read a number of bytes equal to the limit.

ernestask: I don’t have a reliable reproducer, so this isn’t so easy to observe, but the code itself just didn’t make sense to me.

PR #1377, opened 2019-04-15
testsuite: test_order: Add abrt-auto-reporting-sanity-authenticated

ernestask:
It was never added after the split.


ernestask:
Moreover, seems to be failing big time.

mkutlak:
https://github.com/abrt/abrt/blob/14ef09bf801c3c982ea243b1a5a0dc57df6b3d3e/abrt.spec.in#L632

This leads me to believe that this test is not meant for Fedoras.

ernestask:
Yeah: 2f5d47de00ae08e6c22ce3365a8e428cddc3f439. So that’s fine then, we only need to worry about RHEL (where it’s failing). Thanks for pointing it out.

PR #1379, opened 2019-04-29
plugins: Catch unhandled exception in a-a-g-machine-id

mkutlak:
dmidecode can fail due to permission denies or any different reasons that
causes dmidecode to return non-zero return code.

Related: rhbz#1688368

### DIFF
--- a/src/plugins/abrt-action-generate-machine-id
+++ b/src/plugins/abrt-action-generate-machine-id
@@ -23,7 +23,7 @@
 import os
 import sys
 from argparse import ArgumentParser
-from subprocess import check_output
+from subprocess import check_output, CalledProcessError
 import logging
 
 import hashlib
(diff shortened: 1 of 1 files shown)


src/plugins/abrt-action-generate-machine-id
```diff
@@ -52,7 +52,10 @@ def generate_machine_id_dmidecode():
 
     # Run dmidecode command
     for k in keys:
-        data = check_output(["dmidecode", "-s", k]).strip()
+        try:
+            data = check_output(["dmidecode", "-s", k]).strip()
+        except (OSError, CalledProcessError) as ex:
+            raise RuntimeError("Execution of dmidecode failed: {0}".format(str(ex)))
```
mkutlak: Please, suggest a better error message.

ernestask: It doesn’t matter too much. What about including `stdout` or `stderr`?

https://github.com/python/cpython/blob/b0a2c0fa83f9b79616ccf451687096542de1e6f8/Lib/subprocess.py#L112

PR #1383, opened 2019-05-28
daemon: Check return value after reading from channel

### LINKED ISSUE #1168: abrtd: check a return value of g_io_channel_read_chars()
We should check a return value of g_io_channel_read_chars() in function handle_signal_cb() in abrtd.c.

mgrabovsky:
Closes #1168

### DIFF
--- a/src/daemon/abrtd.c
+++ b/src/daemon/abrtd.c
@@ -435,7 +435,24 @@ static gboolean handle_signal_cb(GIOChannel *gio, GIOCondition condition, gpoint
 {
     uint8_t signo;
     gsize len = 0;
-    g_io_channel_read_chars(gio, (void*) &signo, 1, &len, NULL);
+    GError *error = NULL;
+    GIOStatus stat = g_io_channel_read_chars(gio, (void*) &signo, 1, &len, &error);
+    if (stat == G_IO_STATUS_ERROR)
+    {
+        /* An error occurred on the channel. Report it and quit. */
+        error_msg_and_die("Can't read from gio channel: '%s'",
+                error ? error->message : "");
+    }
+    if (stat == G_IO_STATUS_AGAIN)
+    {
+        /* No new data available. Continue as usual. */
+        return G_SOURCE_CONTINUE;
+    }
+    if (stat == G_IO_STATUS_EOF)
+    {
+        /* The channel has reached its end. Remove this event. */
+        return G_SOURCE_REMOVE;
+    }
     if (len == 1)
     {
         /* we did receive a signal */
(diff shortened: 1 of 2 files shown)


src/daemon/abrtd.c
```diff
@@ -435,7 +435,24 @@ static gboolean handle_signal_cb(GIOChannel *gio, GIOCondition condition, gpoint
...
+    if (stat == G_IO_STATUS_ERROR)
+    {
+        /* An error occurred on the channel. Report it and quit. */
+        error_msg_and_die(_("Can't read from gio channel: '%s'"),
+                error ? error->message : "");
+    }
+    if (stat == G_IO_STATUS_AGAIN)
+    {
+        /* No new data available. Continue as usual. */
+        return TRUE;
```
ernestask: Could you use `G_SOURCE_{CONTINUE,REMOVE}` instead? Those are a bit easier to understand at a glance.

src/daemon/abrtd.c
```diff
@@ -435,7 +435,24 @@ static gboolean handle_signal_cb(GIOChannel *gio, GIOCondition condition, gpoint
 {
     uint8_t signo;
     gsize len = 0;
-    g_io_channel_read_chars(gio, (void*) &signo, 1, &len, NULL);
+    GError *error = NULL;
+    GIOStatus stat = g_io_channel_read_chars(gio, (void*) &signo, 1, &len, &error);
+    if (stat == G_IO_STATUS_ERROR)
+    {
+        /* An error occurred on the channel. Report it and quit. */
+        error_msg_and_die(_("Can't read from gio channel: '%s'"),
```
ernestask: Oh, right, we don’t want this to be translatable, since it’s not user-facing.

src/daemon/abrtd.c
```diff
@@ -435,7 +435,24 @@ static gboolean handle_signal_cb(GIOChannel *gio, GIOCondition condition, gpoint
...
+                error ? error->message : "");
```
ernestask: Can the error ever be null if the status indicates there was an error?

mgrabovsky: I believe it is possible, at least in theory. The way some checks in `g_io_channel_read_chars()` are [implemented](https://github.com/GNOME/glib/blob/c8692fffe0723aee2bc6710590b5e70ef841b5cc/glib/giochannel.c#L1992) doesn't seem to preclude `error` being `NULL` even after a failure – for instance, when the channel is not readable. (Which _should not_ happen, but we better be safe than segfault-sorry, right?)

src/daemon/abrtd.c
mgrabovsky: Will fix.

src/daemon/abrtd.c
ernestask: Yeah, that makes sense.

ernestask:
Feel free to merge after pushing.

PR #1387, opened 2019-06-07
Remove default configuration files

mgrabovsky:
https://bugzilla.redhat.com/show_bug.cgi?id=1693287

All the config files were previously installed both to `/usr/share/abrt/conf.d` (defaults, loaded first), as well as `/etc/abrt` (system-wide overloads).

This PR removes the former directory and all its files and preserves the latter. Comments are also added to man pages to document where the files are located.

### DIFF
--- a/doc/abrt-oops.conf.txt
+++ b/doc/abrt-oops.conf.txt
@@ -5,6 +5,10 @@ NAME
 ----
 abrt-oops.conf - Configuration file for ABRT's Kernel Oops extractor
 
+SYNOPSIS
+--------
+/etc/abrt/plugins/oops.conf
+
 DESCRIPTION
 -----------
 The configuration file consists of items in the format "Option = Value".
@@ -17,6 +21,10 @@ DropNotReportableOopses = 'yes' / 'no'
 OnlyFatalMCE = 'yes' / 'no'
    If you want to see only fatal MCEs, set to "yes".
    Defaults is 'yes': detect only fatal ones.
+
+FILES
+-----
+/etc/abrt/plugins/oops.conf
 
 SEE ALSO
 --------
--- a/src/daemon/abrt.conf
+++ b/src/daemon/abrt.conf
@@ -3,47 +3,58 @@
 # Note: you must ensure that whatever directory you specify here exists
 # and is writable for abrtd. abrtd will not create it automatically.
 #
+# Default: none
+#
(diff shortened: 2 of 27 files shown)


doc/abrt-oops.conf.txt
```diff
@@ -5,6 +5,10 @@ NAME
 ----
 abrt-oops.conf - Configuration file for ABRT's Kernel Oops extractor
 
+SYNOPSIS
+--------
+/etc/abrt/plugins/vmcore.conf
```
ernestask: This doesn’t look right.

src/daemon/abrt.conf
```diff
@@ -56,15 +67,18 @@ AutoreportingEnabled = no
 #
 # Caution:
 #
-# THIS FEATURE MIGHT BE USED BY A LOCAL USER TO STEEL YOUR DATA BY ARRANGING A
```
ernestask: At least it can’t be used to plastic my data.

doc/abrt-oops.conf.txt
mgrabovsky: Putting the path into the synopsis? I've sometimes seen that (for instance `resolv.conf`), although I don't have a strong opinion on how "correct" it is. Do you think it would be better to only keep it in the files section at the bottom?

ernestask:
Looks good to me at a glance.

doc/abrt-oops.conf.txt
ernestask: I mean the actual path. It’s not for the same file.

mkutlak:
The `dbus-configuration` test is broken since merge of this PR.

`DefaultConfFile` - still appears on multiple places in the test even thought it was removed from conf files

`DEFAULT_AUTO_REPORTING` is undefined on `line 66`.

mgrabovsky:
Looking into it.

jfilak:
I guess the default files were introduced for Atomic?

jfilak:
Hmm, or the abrt-config daemon was not supposed to touch /usr/share/abrt/conf.d because there is the feature which allows you to reset the configuration to default? If you have only /etc/abrt/conf.d files, then you cannot reset to default.

jfilak:
Aha, so if you drop the configuration DBus daemon, you can also drop the default files :)

PR #1388, opened 2019-06-10
a-a-analyze-c: Fix segfault when saving function name

mgrabovsky:
https://bugzilla.redhat.com/show_bug.cgi?id=1708684

It seems impossible to infer the root cause of the crash from the provided coredump, so this is just a quick fix. This commit only adds a check that `frame` is not `NULL` before dereferencing it to prevent the segfault.

My hunch is that `sr_normalize_core_thread` might get rid of all the stack frames in some corner case. Though, after quickly going through the logic, I'm not sure how that would happen in real world.

### DIFF
--- a/src/plugins/abrt-action-analyze-c.c
+++ b/src/plugins/abrt-action-analyze-c.c
@@ -259,6 +259,12 @@ int main(int argc, char **argv)
         sr_normalize_core_thread(thread);
 
         struct sr_core_frame *frame = thread->frames;
+        if (!frame)
+        {
+            log_info("Could not find any usable stack frame");
+            goto next;
+        }
+
         if (frame->function_name)
             dd_save_text(dd, FILENAME_CRASH_FUNCTION, frame->function_name);
 


ernestask:
> My hunch is that `sr_normalize_core_thread` might get rid of all the stack frames in some corner case. Though, after quickly going through the logic, I'm not sure how that would happen in real world.

Then do we really want to just paint over the potential bug? I needinfo’d the reporter, it could be useful to ask them for the problem directory if it’s always reproducible.

src/plugins/abrt-action-analyze-c.c
```diff
@@ -259,7 +259,7 @@ int main(int argc, char **argv)
         sr_normalize_core_thread(thread);
 
         struct sr_core_frame *frame = thread->frames;
-        if (frame->function_name)
+        if (frame && frame->function_name)
```
ernestask: What if we also printed a message if we have no frames? That would help with future debugging.

mgrabovsky:
I have added a line which logs the failure. Can't think of much else to do right now.

PR #1397, opened 2019-06-19
Revert "applet: Add systemd service unit"

ernestask:
This reverts commit 0ca72ed933dcab065f6f3b741f11fc82820bc835.


ernestask:
Until https://gitlab.gnome.org/GNOME/gnome-session/merge_requests/13 and related work is finished, the service unit won’t work properly. Reverting will also bring back the critical in gnome-session.

ernestask:
Didn’t check if everything works fine, hence the draft.

PR #1400, opened 2019-07-03
spec: Revisit old Python provides

mkutlak:
Time to get rid of these.

Introduced in ab4c89a


xsuchy:
+1

mkutlak:
I moved the abrt-python to python3-abrt changes to https://github.com/abrt/abrt/pull/1401

PR #1406, opened 2019-08-06
Update path to generated file in .gitignore after cli-ng move

mgrabovsky:
This file was moved in #1403 and it was forgotten to update the path here.


ernestask:
An alternative could be adding .gitignore files to each subdirectory, but it might be just as annoying.

Issue #1389, opened 2019-06-11
Lacks bulk delete / remove all / remove all


ringerc:
Both abrt-cli and the GUI lack an obvious bulk-delete or "delete all" option (2.12.0).

That's not super fun when you have a vast number of reports from a kernel oops spamming 10x per second.

I used the GUI and held down the delete key. That worked. An awk script to process the results of `abrt-cli ls` should do the job too.

It looks like the `abrt-cli-ng` tool `abrt` supports `abrt rm '*'` but making this more obvious wouldn't hurt, especially as it doesn't seem to be installed by default (F29).

ringerc:
Correction, `abrt rm '*'` doesn't work.

The docs say

>        MATCH can be either component name, short hash (from list subcommand) or combination of both in form <component>@<hash>

but it looks like it doesn't handle wildcards.

So I landed up writing this incantation to work around it:

    abrt-cli ls | awk '/^Directory:/ { print $2; }' | sudo xargs abrt-cli rm

The new cli appears not to accept multiple arguments to `rm` unlike the old one, which is unfortunate. So while slower, you can use

    sudo abrt ls | awk '/^Id/ { print $2; }' | xargs -n 1 abrt rm

with it.

ernestask:
Resolved with https://github.com/abrt/abrt/pull/1415.

ernestask:
Oh, nevermind, that wasn’t merged.

Issue #1394, opened 2019-06-17
[RFC] Consolidate abrt-cli and abrt-cli-ng


ernestask:
We have two CLI tools for working with problems and both achieve pretty much identical goals, so it might be a good idea to finally drop one of them or combine the good parts of both into a third one.

## Maintainability

`abrt-cli-ng` was written in Python, which, IMO, is going in the right direction as opposed to choosing C. This is a rather uncomplicated tool, but `abrt-cli` even reinvents command-line parsing. On the other hand, `abrt-cli` has better code organization.

## Compatibility

- The console notification script depends on `abrt-cli`, which would call for a port/rewrite.
- `abrt-cli-ng` retains the long forms of the commands/optional arguments, inherited from `abrt-cli` with the exception of `--authenticate`, which became `--auth`.
- `abrt-cli-ng` seems to ship something that looks like an attempt at a bash completion script.
- Both seem to communicate with ABRT over D-Bus for the most part.
- `abrt-cli` allows processing multiple problems interactively

### Related RFEs

https://github.com/abrt/abrt/issues/290
https://github.com/abrt/abrt/issues/291
https://github.com/abrt/abrt/issues/292
https://github.com/abrt/abrt/issues/293
https://github.com/abrt/abrt/issues/631
https://github.com/abrt/abrt/issues/841
https://github.com/abrt/abrt/issues/934
https://github.com/abrt/abrt/issues/1028
https://github.com/abrt/abrt/issues/1279
https://github.com/abrt/abrt/issues/1374
https://github.com/abrt/abrt/issues/1389

### Related BZ bugs

https://bugzilla.redhat.com/show_bug.cgi?id=1329711
https://bugzilla.redhat.com/show_bug.cgi?id=1329712

ernestask:
One idea I had was to make the command-line tool available with the main package and make it properly pluggable with `abrt-action-*` binaries turned into plugins for it (not unlike FAF). That would also help despaghettify the code for some of them.

cmurf:
Top complaints.

1. The trailing / causes an error
[root@flap ~]# abrt-cli report /var/spool/abrt/ccpp-2019-06-19-16\:40\:16.735919-30984/
Can't test whether the element exists over abrt-dbus: GDBus.Error:org.freedesktop.problems.InvalidProblemDir: '/var/spool/abrt/ccpp-2019-06-19-16:40:16.735919-30984/' is not a valid problem directory
[...]

Issue #291, opened 2013-02-04
implement reporting last N crashes


mozeq:
Opened '20110727T08:42:14' by npajkovs as https://fedorahosted.org/abrt/ticket/326

abrt-cli report -N, where N is number

PR #1415, opened 2019-09-05
cli: Tweak problem matching

ernestask:
This commit does several things:
  * Adds a hardcoded “*” pattern for bulk processing
  * Replaces hardcoded “last” pattern with nothing, so that the last
problem is the default
  * Adds path resolution to allow things like relative paths (trailing
slash no longer an issue as well)
  * As a side effect, makes matching against component return all
problems, as opposed to complaining about ambiguity

Related: https://github.com/abrt/abrt/issues/1394

Fixes https://github.com/abrt/abrt/issues/1389
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1329711
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1329712

### DIFF
--- a/src/cli/abrtcli/cli.py
+++ b/src/cli/abrtcli/cli.py
@@ -12,7 +12,7 @@ from reportclient import ask_yes_no, set_verbosity
 
 from abrtcli import config, i18n
 from abrtcli.i18n import _, N_
-from abrtcli.match import match_completer, match_get_problem
+from abrtcli.match import match_completer, match_get_problems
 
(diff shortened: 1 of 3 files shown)


ernestask:
At first, I thought of using regular expressions, but I think it will be less discoverable and just not used by anyone.

ernestask:
/me adjust glasses.

Actually.

I might change the “last” thing again after looking at https://github.com/abrt/abrt/issues/291. We could use `-N 1` everywhere for that.

PR #1418, opened 2019-09-18
spec: Use macros where appropriate

mgrabovsky:
In the process, also fix rpmlint warnings and escape macros in comments.


xsuchy:
LGTM

PR #1419, opened 2019-09-18
Remove abrt-hook-ccpp

ernestask:
With Fedora 26, abrt-ccpp.service is no longer in the system preset and
systemd-coredump took its place:
https://fedoraproject.org/wiki/Changes/coredumpctl

With that, it no longer makes sense to maintain the code, given that no
new user will have it enabled.


ernestask:
This will break integration tests.

ernestask:
The breakage is related to certain tests fiddling with running services (i.e. stopping abrtd also stops abrt-journal-core and then things don’t happen). Hopefully fixed those. Some other ones fail due to recent CLI changes, but I’m cooking up fixes that will go to the other PR.

tests/runtests/docker-essentials/runtest.sh
```diff
@@ -1,123 +0,0 @@
...
-        wait_for_process "abrt-hook-ccpp"
-        rlRun "abrt status > new_status"
-        rlAssertDiffer new_status old_status
-        rlRun "sed -i 's/1/2/' old_status"
-        rlAssertNotDiffer new_status old_status
-
-        rlRun "docker rm abrt_integration_test abrt_integration_test1"
-    rlPhaseEnd
-
-    rlPhaseStartTest "Kill sleep in fedora image with ccpp-plugin"
```
mkutlak: Any reason to remove the whole test instead of just this `ccpp-hook` part?


I think you can remove the `wait_for_process "abrt-hook-ccpp"` on line 74 as it looks to me as a Copy Pasta mysteak.

mkutlak:
Other than that it looks good to me.


Au revoir abrt-hook-ccpp. :wave:

tests/runtests/docker-essentials/runtest.sh
ernestask: I removed everything that looked like testing `abrt-hook-ccpp` exclusively, so probably just an oversight from my end.

tests/runtests/aux/test_order
```diff
@@ -24,24 +24,13 @@ abrt-crash-unpackaged
...
 ccpp-plugin-java
-ccpp-plugin-config
-ccpp-plugin-hook-unwind
-ccpp-plugin-selinux
-ccpp-plugin-debug
-ccpp-plugin-core-size
 python3-addon
 
-# - containers
-docker-essentials
```
mkutlak: Keep this to run the test. (same in test_order.rhel8)

mkutlak:
@trapas retest this please

PR #1422, opened 2019-10-16
Use correct paths for installed Python modules

mgrabovsky:
According to the [packaging guide](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_macros), we should be using `%{python3_sitelib}` to install Python modules rather than `%{python3_sitearch}` which is more suitable for compiled extensions. This also needs to be reflected in the automake scripts, where we need to replace `$(pyexecdir)` with `$(pythondir)`. (See [automake manual](https://www.gnu.org/software/automake/manual/html_node/Python.html).)

This is a follow-up to 20dcf7fb4.

### DIFF
--- a/abrt.spec.in
+++ b/abrt.spec.in
@@ -935,14 +935,14 @@ killall abrt-dbus >/dev/null 2>&1 || :
 %{_journalcatalogdir}/python3_abrt.catalog
 %config(noreplace) %{_sysconfdir}/libreport/plugins/catalog_python3_format.conf
 %{_mandir}/man5/python3_event.conf.5*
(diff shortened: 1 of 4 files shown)


ernestask:
And what about the changes in build files? How do those differ, exactly?

ernestask:
Right, so https://www.gnu.org/software/automake/manual/html_node/Python.html says that `pyexecdir` is for shared libraries.

mgrabovsky:
Yes, yes, I'll update the commit message.

mgrabovsky:
Updated the commit message and description. (Sorry for the delay, I was having my vision checked.)

PR #1423, opened 2019-10-19
Bump GLib dependency version to >=2.55.1

mattst88:
g_application_set_option_context_summary was introduced in glib 2.55.1.

Bug: https://bugs.gentoo.org/682840


trapas:

ernestask:
Oops, thanks.

PR #1425, opened 2019-11-08
cli: Print usage if non-action option passed without a command

ernestask:
See commits.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1770034


mgrabovsky:
Looks good 👍

PR #1429, opened 2020-01-14
Use GChecksum to compute checksums

ernestask:
d605ffeaa6ae411ef396160ffd67b7a6fd27c6ba in libreport removed some
convenience API for hashing data, so let’s just go with what GLib has in
store.


ernestask:
I see identical checksums being generated with this, so seems like it does the trick.

mkutlak:
LGTM

PR #1430, opened 2020-01-15
Move from nose to pytest

ernestask:
The original has seen its last commit in 2016. While nose2 exists,
pytest allows writing tests using the assert statement without losing
context.


ernestask:
Moreover, python-nose might disappear from Fedora: https://lists.fedorahosted.org/archives/list/devel@lists.fedoraproject.org/thread/NNGYNRUH5VBA2MO57C3VE3ON4QFAPEKR/

src/python-problem/tests/Makefile.am
```diff
@@ -4,11 +4,8 @@ check_SCRIPTS = $(TESTS)
...
-check-local:
-	echo "Running tests with python3"; \
-	exc=0; \
-	for test in $(TESTS); do \
-		python3 $$test || exc=1; \
-	done; \
-	exit $$exc
+if HAVE_PYTEST
+check-local: $(TESTS)
+	pytest-3 --cov=problem --capture=no --verbose $(TESTS)
```
mkutlak:
```
ERROR: usage: pytest-3 [options] [file_or_dir] [file_or_dir] [...]                                                     
pytest-3: error: unrecognized arguments: --cov=problem                                                                                                                                                                                        
  inifile: None                                                                                                        
  rootdir: /home/mkutlak/ABRT/abrt/build/abrt-2.12.2.67.gb050/src/python-problem/tests
```

ernestask: Ah, good catch, I think it’s a plugin. Better to not care about coverage here, then, but rather in Jenkins.

mkutlak:
image

LGTM

PR #1434, opened 2020-01-23
koops: Avoid explicit dependency on hash size

mgrabovsky:
Change the API so that `koops_hash_str` and `koops_hash_str_ext` return the heap-allocated encoded digest instead of taking a pointer and returning a status code.

This eliminates the need to know the size of the resulting hash in advance, which should solve the build problems following recent libreport changes.


ernestask:
Awesome, thanks.

mkutlak:
These changes (https://github.com/abrt/abrt/pull/1429, https://github.com/abrt/libreport/pull/588) require a bump of 
https://github.com/abrt/abrt/blob/4a8cadcde475a50f0194c25bab1d31e08ac17345/abrt.spec.in#L46
during next `abrt - libreport - satyr` release.


Otherwise `abrt` becomes broken. (Experiencing this on my `rawhide` machine.)

PR #1436, opened 2020-01-28
Fix build failure with gcc -fno-common

michalfabik:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1795820

Signed-off-by: Michal Fabik mfabik@redhat.com

### DIFF
--- a/src/dbus/abrt_problems2_service.h
+++ b/src/dbus/abrt_problems2_service.h
@@ -86,14 +86,14 @@ GVariant *abrt_p2_service_entry_problem_data(AbrtP2Service *service,
             GError **error);
 
 
-enum {
+typedef enum {
     ABRT_P2_SERVICE_ENTRY_LOOKUP_NOFLAGS  = 0x0, ///< return with error if not found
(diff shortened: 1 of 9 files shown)


src/dbus/abrt_problems2_service.h
```diff
@@ -86,7 +86,7 @@ GVariant *abrt_p2_service_entry_problem_data(AbrtP2Service *service,
             GError **error);
 
 
-enum {
+extern enum {
```
ernestask: We don’t need this at all. A better thing to do would be to `typedef` it and rename to `AbrtP2ServiceEntryLookupFlags` or something like that and then use that instead of plain `int` in parameters/variables.

ernestask:
https://bugzilla.redhat.com/show_bug.cgi?id=1795820 should also be referenced now.

src/dbus/abrt_problems2_service.h
michalfabik: Changed.

PR #1438, opened 2020-01-29
dbus: Warn about large number of problems

ernestask:
While enumerating all the problems in itself might not be too slow,
requesting properties for each *will* be slow. This commit adds a
warning to the user that a cleanup might be in order.


ernestask:
:man_shrugging:

mgrabovsky:
Seems reasonable.

PR #1440, opened 2020-01-30
bodhi: Initialize karma values with default

mkutlak:
During a rebuilt on s390x this error came up:
```
bodhi.c:340:12:
  error:
    'unstable_karma' may be used uninitialized in this function
```

### DIFF
--- a/src/plugins/bodhi.c
+++ b/src/plugins/bodhi.c
@@ -334,7 +334,8 @@ static GHashTable *bodhi_parse_json(json_object *json, const char *release)
         if (!builds_item) /* broken json */
             continue;
 
-        int karma, unstable_karma;
+        int karma = 0;
+        int unstable_karma = 0;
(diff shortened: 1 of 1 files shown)


mkutlak:
https://kojipkgs.fedoraproject.org//work/tasks/3715/41253715/build.log

src/plugins/bodhi.c
```diff
@@ -335,6 +335,7 @@ static GHashTable *bodhi_parse_json(json_object *json, const char *release)
             continue;
 
         int karma, unstable_karma;
+        karma = unstable_karma = 0;
```
ernestask: Please, *please*, **please** something like this, instead:

```c
int karma = 0;
int unstable_karma = 0;
```

mkutlak: sorry :roll_eyes:

ernestask: karma++

PR #1443, opened 2020-02-07
spec: bump reqired libreport version

mkutlak:
New libreport changes are incompatible with older abrt releases.


ernestask:
Argh, forgot about it.

PR #1448, opened 2020-02-19
a-a-save-package-data: Use regexps to match interpreters

mkutlak:
Instead of adding more and more interpreters to the list which
gets outdated after a while, we can utilize regular expressions.

This removes the configurable option for Interpreters from
/etc/abrt/abrt-action-save-package-data.conf


src/daemon/abrt-action-save-package-data.c
```diff
@@ -25,15 +26,18 @@
...
 #define DEFAULT_BLACKLISTED_PKGS "bash, mono-core, nspluginwrapper, strace, valgrind"
 #define DEFAULT_GPG_KEYS_DIR "/etc/pki/rpm-gpg"
-#define DEFAULT_INTERPRETERS "python, python2, python2.7, python3, python3.3, " \
-    "python3.4, python3.5, python3.6, python3.7, perl, perl5.16.2"
+#define DEFAULT_INTERPRETERS_REGEX \
+    "(perl ([:digit:][.][:digit:]+[.][:digit:])?" \
+    "|" \
+    "php" \
+    "|" \
+    "(platform-)? python ([:digit:]([.][:digit:])?)?)$"
```
mgrabovsky: For some reason, the regex doesn't work for me with `[:digit:]` but it does with `\d` (tried on a trivial harness).

mgrabovsky: Oh right, it should be `[[:digit:]]` (with the two pairs brackets).

ernestask:
Why get rid of the configuration setting, though?

src/daemon/abrt-action-save-package-data.c
mkutlak: Yeah, I used `\d` at first, but then in an attempt to make the regex more readable I've changed it to `:digit:` :man_facepalming:

Fixed.

doc/abrt-action-save-package-data.conf.txt
```diff
@@ -42,8 +42,7 @@ A description of each item follows:
    a program whose basename is equal to one of these, ABRT assigns the crash to the
    executed script instead.
    +
-   Default is `python, python2, python2.7, python3, python3.3, python3.4, python3.5,
-   python3.6, python3.7, perl, perl5.16.2`.
+   Default is `python, perl, php`.
```
mkutlak: I am not sure what to put here. `php, perl, python2, python3`?

ernestask: I imagine we would need to hack on support for multiple configuration values, e.g.

```
Interpreters=(perl ([[:digit:]][.][[:digit:]]+[.][[:digit:]])?
Interpreters=php
Interpreters=(platform-)? python ([[:digit:]]([.][[:digit:]])?)?)$
```

And then you could probably just write it as a list.

mkutlak:
> Why get rid of the configuration setting, though?

I brought it back. 

My initial reasoning was that this (don't blame the interpreter) should be for the maintainers of the interpreters, so they don't get assigned bugs/crashes from other components.

doc/abrt-action-save-package-data.conf.txt
mkutlak: Meh, that looks ugly and complicated.

ernestask:
Sure, but we’d need to rebuild the binary for every change. Now, we can ship a config override with the package or simply document the setting in case we have someone developing something locally and using ABRT.

/me shrugs.

doc/abrt-action-save-package-data.conf.txt
ernestask: It’s as ugly and as complicated as the regular expressions are, really, but I see your point.

ernestask: Okay, counter-proposal(s):

1. Hardcode meta-values, like `@perl` and `@python` and reference those in a comma-separated list, like before
2. Allow to configure such meta-values in the config (allowing regular expressions) and do 1

src/daemon/abrt-action-save-package-data.c
```diff
@@ -320,7 +323,8 @@ static int SavePackageDescriptionToDebugDump(const char *dump_dir_name, const ch
     /* if basename is known interpreter, we want to blame the running script
      * not the interpreter
      */
-    if (g_list_find_custom(settings_Interpreters, basename, (GCompareFunc)g_strcmp0))
+    if (g_regex_match_simple(DEFAULT_INTERPRETERS_REGEX, basename, G_REGEX_EXTENDED, /*MatchFlags*/0) ||
+        g_list_find_custom(settings_Interpreters, basename, (GCompareFunc)g_strcmp0))
```
ernestask: Is the cast for `g_strcmp0()` really needed? I forget my C-fu, but the signatures should be compatible here.

mkutlak:
```
abrt-action-save-package-data.c: In function 'SavePackageDescriptionToDebugDump':
abrt-action-save-package-data.c:327:61: error: passing argument 3 of 'g_list_find_custom' from incompatible pointer type [-Werror=incompatible-pointer-types]
  327 |         g_list_find_custom(settings_Interpreters, basename, g_strcmp0))
      |                                                             ^~~~~~~~~                   
      |                                                             |                                
      |                                                             int (*)(const char *, const char *)
In file included from /usr/include/glib-2.0/glib/ghash.h:33,
                 from /usr/include/glib-2.0/glib.h:50,                                                                 
                 from abrt-action-save-package-data.c:20:
/usr/include/glib-2.0/glib/glist.h:122:25: note: expected 'GCompareFunc' {aka 'int (*)(const void *, const void *)'} but argument is of type 'int (*)(const char *, const char *)'
  122 |       GCompareFunc      func);
      |       ~~~~~~~~~~~~~~~~~~^~~~                                                                                   
cc1: all warnings being treated as errors
```

jfilak:
Just for the record, @mhabrnal already attempted to turn the option Interpreters to a regex but I asked him to keep it a simple list because I thought it is more reliable as we have to deal with potentially complex problem which could lead into un-maintainable regex providing unexpected results. This is just to let you know about our former decisions and I do not attempt to convince you to keep it as a pure list.

I am just not sure why you removed the option Interprets from the tests and I don't see any new test validating all the regexp parts works as expected.

It would be also nice to write down the list of cases you wanted to cover by the regex.

tests/runtests/dont-blame-interpret/PURPOSE
```diff
@@ -1,3 +1,3 @@
 PURPOSE of dont-blame-interpret
-Description: in case of crash in interpret (python, perl) abrt should blame the running script not the interpret issue#609
+Description: in case of crash in interpret (python, perl, R, tcl, php) abrt should blame the running script not the interpret issue#609
```
ernestask: I guess this incoherent mess of a sentence can also be fixed while we’re at it.

mkutlak: Changed. Not sure if it's better.

ernestask: Verily better.

mkutlak:
> Just for the record, @mhabrnal already attempted to turn the option Interpreters to a regex but I asked him to keep it a simple list because I thought it is more reliable as we have to deal with potentially complex problem which could lead into un-maintainable regex providing unexpected results. This is just to let you know about our former decisions and I do not attempt to convince you to keep it as a pure list.

Thank you for this. 

I feel like with the list there is always a question `When is it OK to remove XZ interpreter?`. 
Another thing is keeping track of new interpreter releases.

> I am just not sure why you removed the option Interprets from the tests and I don't see any new test validating all the regexp parts works as expected.

Reverted and added new tests.

PR #1464, opened 2020-03-25
plugins: abrt-journal-core: Don’t assume anything about uid_t

ernestask:
Sure, maybe glibc defines uid_t as a uint32_t, but POSIX does not impose
any limits on the underlying type, so let’s just accept whatever systemd
gives us. If what systemd gives us is non-compliant, then we are already
in trouble and it makes no sense to defend against nasal demons.
Moreover, using a long for bounds and passing UINT_MAX is just asking
for trouble (i.e. it fails on 32-bit systems).


ernestask:
This makes abrt-dump-journal-core usable on 32-bit systems.

PR #1465, opened 2020-03-26
Hardcode subcomponents for RHBZ reports

michalfabik:
This PR adds default (for lack of a better word) subcomponents for components that require them, as offered in the BZ form for submitting a new bug. The subcomponents are stored in `sub_components` file in the problem directory, to be read by `reporter-bugzilla` (see https://github.com/abrt/libreport/pull/608).  
  
(Not quite sure where to put the function since unlike `rpm_get_component()`, it doesn't rely on rpm's in any way, so putting it in `rpm.c` didn't really feel appropriate.)

### DIFF
--- a/src/daemon/abrt-action-save-package-data.c
+++ b/src/daemon/abrt-action-save-package-data.c
@@ -273,6 +273,7 @@ static int SavePackageDescriptionToDebugDump(const char *dump_dir_name, const ch
     char *fingerprint = NULL;
     struct pkg_envra *pkg_name = NULL;
     char *component = NULL;
+    char *sub_components = NULL;
     char *kernel = NULL;
     int error = 1;
     /* note: "goto ret" statements below free all the above variables,
@@ -411,6 +412,7 @@ skip_interpreter:
     }
 
     component = rpm_get_component(executable, chroot);
+    sub_components = get_sub_components(component);
 
     dd = dd_opendir(dump_dir_name, /*flags:*/ 0);
     if (!dd)
@@ -449,6 +451,8 @@ skip_interpreter:
 
     if (component)
         dd_save_text(dd, FILENAME_COMPONENT, component);
+    if (sub_components)
+        dd_save_text(dd, FILENAME_SUBCOMPONENTS, sub_components);
 
  ret0:
     error = 0;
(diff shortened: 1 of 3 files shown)


src/include/libabrt.h
```diff
@@ -52,6 +52,8 @@ void ensure_writable_dir_group(const char *dir, mode_t mode, const char *user, c
 char *run_unstrip_n(const char *dump_dir_name, unsigned timeout_sec);
 #define get_backtrace abrt_get_backtrace
 char *get_backtrace(const char *dump_dir_name, unsigned timeout_sec, const char *debuginfo_dirs);
+#define get_sub_components abrt_get_sub_components
```
ernestask: Please don’t do this. Just declare the function as `abrt_get_sub_components()`, otherwise it will be impossible to debug.

src/lib/hooklib.c
```diff
@@ -415,6 +415,28 @@ char *get_backtrace(const char *dump_dir_name, unsigned timeout_sec, const char
     return bt;
 }
 
+char *get_sub_components(const char *component)
+{
+    if (!strcmp(component, "binutils"))      return (char *)"system-version";
```
ernestask: I think it would be better to define an array and loop over it.

src/daemon/abrt-action-save-package-data.c
```diff
@@ -449,6 +451,8 @@ static int SavePackageDescriptionToDebugDump(const char *dump_dir_name, const ch
 
     if (component)
         dd_save_text(dd, FILENAME_COMPONENT, component);
+    if (sub_components)
+        dd_save_text(dd, FILENAME_SUBCOMPONENTS, sub_components);
```
ernestask: Again, what is the point? The only time you need to know the subcomponents are when sending the report. I.e. you do just-in-time negotiation.

src/lib/hooklib.c
ernestask: And you need a default regardless.

ernestask:
I think this is the wrong place for these changes. It should strictly be part of the reporter.

src/include/libabrt.h
michalfabik: That's what I'm thinking every time I run into this. What is the point of these `#define`s exactly?  
  
Anyway, I'll move the whole thing to libreport, I'm just asking generally.

ernestask: I have no idea, but I hate it every time when gdb doesn’t understand me when I try to add a breakpoint.

michalfabik: Wouldn't it make sense to get rid of those altogether then?

ernestask: Yeah, I think their only purpose is to be gotten rid of. If you’re sufficiently tickled by this, I absolutely welcome such changes.

PR #1466, opened 2020-03-26
Remove #define's adding abrt_ to function names

michalfabik:
See commit, per @ernestask 's suggestion. I couldn't find an obvious reason for these with blame, so unless it's a case of some clever (ob)fuscery like assembling function names out of substrings before calling them, this should be safe.

### DIFF
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -253,7 +253,7 @@ get_autoreport_event_name (void)
         return configured;
     }
 
-    return g_settings_autoreporting_event;
+    return abrt_g_settings_autoreporting_event;
 }
 
 static bool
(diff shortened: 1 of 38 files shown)


ernestask:
The idea behind them is most likely adding a namespace without having to type too much later, so the proper way to go would be to subsequently namespace all those functions and globals and change all calls (it’s especially bad with the `g_settings_*` variables, because it could potentially clobber GLib code.

michalfabik:
Ah, so, like this? (see update)

ernestask:
Yup, that’s what I had in mind. Thanks and sorry for the manual labor, but it’s badly needed.

michalfabik:
Not at all! And to open this can of worms all the way, I have a similar PR ready for libreport _and_ then again for ABRT to fix the libreport calls that get broken by the former.  
  
(I haven't opened the PRs yet in fear someone might have a fit when they see the gazillion changes all at once.)

PR #1477, opened 2020-04-23
applet: Refactor code

ernestask:
See commits.


ernestask:
Should help with https://bugzilla.redhat.com/show_bug.cgi?id=1812488, but I’m not sure how to reproduce the crash.

ernestask:
I found some more opportunities to clean up, so this will take a bit longer.

ernestask:
Okay, this ended up being a bit more invasive than originally planned.

PR #1479, opened 2020-04-29
Replace various helper functions with GLib ones

michalfabik:
See commits. See also abrt/libreport#636 and https://github.com/abrt/abrt-java-connector/pull/10

### DIFF
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -196,7 +196,7 @@ static void
 migrate_auto_reporting_to_gsettings (void)
 {
 #define OPT_NAME "AutoreportingEnabled"
-    map_string_t *settings_map;
+    GHashTable *settings_map;
     int sv_logmode;
     int auto_reporting;
     int configured;
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -37,7 +37,7 @@ typedef struct
 {
     pid_t child_pid;
     int child_stdout_fd;
-    struct strbuf *cmd_output;
+    GString *cmd_output;
 
     AbrtAppletApplication *application;
     AbrtAppletProblemInfo *problem_info;
@@ -53,7 +53,7 @@ event_processing_state_new (void)
 
     state->child_pid = -1;
     state->child_stdout_fd = -1;
-    state->cmd_output = libreport_strbuf_new ();
+    state->cmd_output = g_string_new(NULL);
 
     return state;
 }
@@ -61,7 +61,7 @@ event_processing_state_new (void)
 static void
 event_processing_state_free (EventProcessingState *state)
 {
-    libreport_strbuf_free (state->cmd_output);
+    g_string_free(state->cmd_output, TRUE);
     g_clear_object (&state->problem_info);
     g_clear_object (&state->application);
 
@@ -808,17 +808,17 @@ handle_event_output_cb (GIOChannel   *gio,
         {
             *newline = '\0';
 
-            libreport_strbuf_append_str (state->cmd_output, raw);
+            g_string_append(state->cmd_output, raw);
 
-            log_debug ("%s", state->cmd_output->buf);
+            log_debug ("%s", state->cmd_output->str);
 
-            libreport_strbuf_clear(state->cmd_output);
+            g_string_erase(state->cmd_output, 0, -1);
             /* jump to next line */
             raw = newline + 1;
         }
 
         /* beginning of next line. the line continues by next read */
-        libreport_strbuf_append_str (state->cmd_output, raw);
+        g_string_append(state->cmd_output, raw);
     }
 
     /* EOF/error */
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -196,16 +196,14 @@ static void
 migrate_auto_reporting_to_gsettings (void)
 {
 #define OPT_NAME "AutoreportingEnabled"
-    GHashTable *settings_map;
+    g_autoptr(GHashTable) settings_map = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
     int sv_logmode;
     int auto_reporting;
     int configured;
 
-    settings_map = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
-
     if (!libreport_load_app_conf_file (APP_NAME, settings_map))
     {
-        goto finito;
+        return;
     }
 
     /* Silently ignore not configured options */
@@ -220,7 +218,7 @@ migrate_auto_reporting_to_gsettings (void)
 
     if (configured == 0)
     {
-        goto finito;
+        return;
     }
 
     /* Enable the GS option if AutoreportingEnabled is true because the user
@@ -245,9 +243,6 @@ migrate_auto_reporting_to_gsettings (void)
     log_warning ("Successfully migrated "APP_NAME":"OPT_NAME" to "GS_SCHEMA_ID_PRIVACY":"GS_PRIVACY_OPT_AUTO_REPORTING);
 
 #undef OPT_NAME
-finito:
-    if (settings_map)
-        g_hash_table_destroy(settings_map);
 }
 
 static const char *
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -298,13 +298,13 @@ new_dir_exists (GList **new_dirs)
     }
 
     cachedir = g_get_user_cache_dir ();
-    dirlist_name = libreport_concat_path_file (cachedir, "abrt");
+    dirlist_name = g_build_filename(cachedir, "abrt", NULL);
 
     g_mkdir_with_parents (dirlist_name, 0777);
 
     g_free (dirlist_name);
 
-    dirlist_name = libreport_concat_path_file (cachedir, "abrt/applet_dirlist");
+    dirlist_name = g_build_filename(cachedir, "abrt/applet_dirlist", NULL);
     fp = fopen (dirlist_name, "r+");
     if (fp == NULL)
     {
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -201,7 +201,7 @@ migrate_auto_reporting_to_gsettings (void)
     int auto_reporting;
     int configured;
 
-    settings_map = libreport_new_map_string ();
+    settings_map = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
 
     if (!libreport_load_app_conf_file (APP_NAME, settings_map))
     {
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -351,7 +351,7 @@ new_dir_exists (GList **new_dirs)
             {
                 if (new_dirs != NULL)
                 {
-                    *new_dirs = g_list_prepend (*new_dirs, libreport_xstrdup (l1->data));
+                    *new_dirs = g_list_prepend (*new_dirs, g_strdup (l1->data));
 
                     log_notice ("New dir detected: %s", (char *) l1->data);
                 }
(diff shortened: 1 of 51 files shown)


ernestask:
One of the commits includes changes to abrt.pot. Those should be dropped.

src/configuration-gui/abrt-config-widget.c
```diff
@@ -120,9 +120,9 @@ static void abrt_config_widget_finalize(GObject *object);
 static AbrtAppConfiguration *
 abrt_app_configuration_new(const char *app_name)
 {
-    AbrtAppConfiguration *conf = libreport_xmalloc(sizeof(*conf));
+    AbrtAppConfiguration *conf = g_malloc(sizeof(*conf));
```
ernestask, line 123:
```suggestion
    AbrtAppConfiguration *conf = g_new0(AbrtAppConfiguration, 1);
```

src/plugins/abrt-dump-oops.c
```diff
@@ -17,6 +17,7 @@
        Arjan van de Ven <arjan@linux.intel.com>
  */
 #include <syslog.h>
+#include <glib-2.0/glib/gstdio.h>
```
ernestask, line 20:
```suggestion
#include <glib/gstdio.h>
```

michalfabik:
> One of the commits includes changes to abrt.pot. Those should be dropped.  
  
Two actually. Fixed, thanks.

src/plugins/abrt-dump-oops.c
michalfabik: Fixed.

src/configuration-gui/abrt-config-widget.c
michalfabik: Addressed in a new commit for all other instances of the same.

michalfabik:
This is broken, a double `free()` in `abrtd.service`. Investigating now.

src/dbus/abrt-dbus.c
```diff
@@ -376,29 +370,26 @@ static void handle_method_call(GDBusConnection *connection,
...
         g_dbus_method_invocation_return_value(invocation, response);
-        free(problem_id);
 
         return;
     }
 
     if (g_strcmp0(method_name, "GetProblems") == 0)
     {
-        GList *dirs = get_problem_dirs_for_uid(caller_uid, abrt_g_settings_dump_location);
+        g_autolist (GList) dirs = get_problem_dirs_for_uid(caller_uid, abrt_g_settings_dump_location);
```
ernestask, line 391: Uhh… You are declaring a list of lists here.

src/plugins/abrt-dump-xorg.c
```diff
@@ -16,6 +16,7 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+#include <glib-2.0/glib/gstdio.h>
```
ernestask, line 19: This is not what the path should be. You got it right (almost) everywhere else, so I imagine you just need to adjust the build definitions.

src/plugins/bodhi.c
```diff
@@ -181,7 +181,7 @@ static void bodhi_read_value(json_object *json, const char *item_name,
         *(int *) value = json_object_get_int(j);
         break;
     case BODHI_READ_STR:
-        *(char **) value = (char *) libreport_strtrimch(libreport_xstrdup(json_object_to_json_string(j)), '"');
+        *(char **) value = (char *) libreport_strtrimch(g_strdup(json_object_to_json_string(j)), '"');
```
ernestask, line 184: `libreport_strtrimch()` could possibly be replaced with `g_strdelimit()`.

src/plugins/bodhi.c
```diff
@@ -538,41 +535,40 @@ int main(int argc, char **argv)
             if (!problem_data)
                 libreport_xfunc_die(); /* create_problem_data_for_reporting already emitted error msg */
 
-            char *product, *version;
-            map_string_t *osinfo = libreport_new_map_string();
+            g_autofree char *product = NULL;
+            g_autofree char *version = NULL;
+            map_string_t *osinfo = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
```
ernestask, line 540: You could probably opt for automatic cleanup (`map_string_t` is just a typedef for `GHashTable`).

src/plugins/bodhi.c
```diff
@@ -513,9 +510,9 @@ int main(int argc, char **argv)
     if (!bugs && !argv[optind])
         libreport_show_usage_and_die(program_usage_string, program_options);
 
-    struct strbuf *query = libreport_strbuf_new();
+    GString *query = g_string_new(NULL);
```
ernestask, line 513: You can also get automatic cleanup for `GString` (https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/glib-autocleanups.h).

src/plugins/oops-utils.c
```diff
@@ -37,12 +37,11 @@ int abrt_oops_process_list(GList *oops_list, const char *dump_location, const ch
             int i = 0;
             while (i < oops_cnt)
             {
-                char *kernel_bt = (char*)g_list_nth_data(oops_list, i++);
-                char *tainted_short = abrt_kernel_tainted_short(kernel_bt);
+                g_autofree char *kernel_bt = (char*)g_list_nth_data(oops_list, i++);
```
ernestask, line 40: I don’t see this freed in the old code.

tests/examples/oops_full_hostname.test
```diff
@@ -1 +1 @@
-Oct 12 23:25:16 coprbox.den kernel: mce: [Hardware Error]: Machine check events logged
+Oct 12 23:25:16 ishap kernel: mce: [Hardware Error]: Machine check events logged
```
ernestask, line 1: This needs to be dropped.

src/applet/abrt-applet-application.c
```diff
@@ -201,7 +201,7 @@ migrate_auto_reporting_to_gsettings (void)
     int auto_reporting;
     int configured;
 
-    settings_map = libreport_new_map_string ();
+    settings_map = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
```
mgrabovsky, line 204: This could also be a `g_autoptr(GHashTable)`.

src/configuration-gui/abrt-config-widget.c
```diff
@@ -120,10 +120,10 @@ static void abrt_config_widget_finalize(GObject *object);
 static AbrtAppConfiguration *
 abrt_app_configuration_new(const char *app_name)
 {
-    AbrtAppConfiguration *conf = libreport_xmalloc(sizeof(*conf));
+    AbrtAppConfiguration *conf = g_new(AbrtAppConfiguration, 1);
 
-    conf->app_name = libreport_xstrdup(app_name);
-    conf->settings = libreport_new_map_string();
+    conf->app_name = g_strdup(app_name);
```
mgrabovsky, line 125: Is it okay to replace `libreport_xstrdup` with `g_strdup`? The former aborts with a message if there's not enough memory, the latter only returns `NULL`. I think we should at least check the return value if we're going to use the GLib functions here. (Same with most of the `libreport_x*` functions.)

src/daemon/abrt-action-save-package-data.c
```diff
@@ -69,67 +69,66 @@ static GList *settings_Interpreters = NULL;
...
     {
-        error_msg("Unrecognized variable '%s' in '%s'", name, conf_filename);
+        error_msg("Unrecognized variable '%s' in '%s'", (char *)name, conf_filename);
     }
 }
 
 static void load_gpg_keys(void)
 {
-    map_string_t *settings = libreport_new_map_string();
+    map_string_t *settings = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
```
mgrabovsky, line 124: This variable doesn't seem to be freed anywhere inside the function. Another candidate for autoptr.

src/plugins/abrt-action-analyze-python.c
```diff
@@ -62,10 +62,10 @@ int main(int argc, char **argv)
     struct dump_dir *dd = dd_opendir(dump_dir_name, /*flags:*/ 0);
     if (!dd)
         return 1;
-    char *bt = dd_load_text(dd, FILENAME_BACKTRACE);
+    g_autofree char *bt = dd_load_text(dd, FILENAME_BACKTRACE);
```
mgrabovsky, line 65: Leftover `free(bt)` on line 97. (Apparently can't add a comment there.)

src/plugins/bodhi.c
```diff
@@ -584,27 +580,25 @@ int main(int argc, char **argv)
...
-        libreport_strbuf_append_strf(q, " %s", b->nvr);
+        g_string_append_printf(q, " %s", b->nvr);
     }
 
     /*g_hash_table_unref(update_hash_tbl);*/
 
     if (!q->len)
     {
-        /*libreport_strbuf_free(q);*/
+        /*g_string_free(q, TRUE);*/
```
mgrabovsky, line 601: Could lose these two "comments" altogether.

src/plugins/bodhi.c
```diff
@@ -621,24 +615,25 @@ int main(int argc, char **argv)
...
+    if (settings)
+        g_hash_table_destroy(settings);
+
+    char *msg = g_strdup_printf(_("An update exists which might fix your problem. "
+                                  "You can install it by running: %s. "
+                                  "Do you want to continue with reporting the bug?"),
+                                q->str
     );
-    /*libreport_strbuf_free(q);*/
+    /*g_string_free(q, TRUE);*/
```
mgrabovsky, line 636: Drop this.

src/dbus/abrt_problems2_service.c
```diff
@@ -1125,11 +1122,10 @@ static GVariant *entry_object_dbus_get_property(GDBusConnection *connection,
         GVariantBuilder builder;
         g_variant_builder_init(&builder, G_VARIANT_TYPE("as"));
         dd_init_next_file(dd);
-        char *short_name;
+        g_autofree char *short_name = NULL;
         while (dd_get_next_file(dd, &short_name, NULL))
         {
             g_variant_builder_add(&builder, "s", short_name);
-            free(short_name);
         }
```
mgrabovsky, line 1129: I don't think this is equivalent. In the original code, `short_name` was freed in every iteration, now only the last value is freed once it goes out of scope, so there will be a leak.

src/lib/abrt_conf.c
```diff
@@ -134,35 +134,34 @@ static void ParseCommon(map_string_t *settings, const char *conf_filename)
...
     }
 
     GHashTableIter iter;
-    const char *name;
-    /*char *value; - already declared */
-    libreport_init_map_string_iter(&iter, settings);
-    while (libreport_next_map_string_iter(&iter, &name, &value))
+    gpointer name;
+    g_hash_table_iter_init(&iter, settings);
+    while (g_hash_table_iter_next(&iter, &name, &value))
```
mgrabovsky, line 162: The third argument can be `NULL` since we don't use it.

src/configuration-gui/abrt-config-widget.c
ernestask: It is, GLib memory-allocation functions also abort.

ernestask: There’s even a recent commit to clarify this: https://gitlab.gnome.org/GNOME/glib/-/commit/445aa65e900cc2c0a861739b660f00d0b748a6f6.

mgrabovsky: Ah, nice, didn't know that. I tried to follow the path down to the `_GLIB_NEW` macro but didn't find any mention of aborting.

ernestask: For science:
https://gitlab.gnome.org/GNOME/glib/-/blob/c93318953eec6a65de3e466f648a4f26109ff0eb/glib/gmem.h#L283 translates to `g_malloc()` being called, which has https://gitlab.gnome.org/GNOME/glib/-/blob/c93318953eec6a65de3e466f648a4f26109ff0eb/glib/gmem.c#L111 and which eventually ends up at https://gitlab.gnome.org/GNOME/glib/-/blob/c93318953eec6a65de3e466f648a4f26109ff0eb/glib/gmessages.c#L556.

mgrabovsky:
Three unresolved comments still left to be fixed.

Issue #1357, opened 2019-01-17
ABRT deletes core dumps far too aggressively, even while user is trying to report the crash


mcatanzaro:
ABRT has failed to detect any crashes for me for several months now, maybe longer. No crashes appear in ABRT's GUI, even though crashes happen. Running `journalctl -u` I see various errors from abrt whenever a crash occurs. Here's a week's worth of errors:

```
Jan 12 16:51:56 chargestone-cave abrtd[3593]: Size of '/var/spool/abrt' >= 5000 MB (MaxCrashReportsSize), deleting old directory 'ccpp-2019-01-11-12:08:59.550787-127263'
Jan 12 16:51:56 chargestone-cave abrtd[3593]: Size of '/var/spool/abrt' >= 5000 MB (MaxCrashReportsSize), deleting old directory 'Python3-2019-01-07-22:06:40-54524'
Jan 12 16:51:56 chargestone-cave abrt-server[111301]: Unsupported container technology
Jan 12 16:51:56 chargestone-cave abrt-server[111301]: Lock file '.lock' was locked by process 111304, but it crashed?
Jan 12 16:51:56 chargestone-cave abrt-server[111301]: Error: No segments found in coredump './coredump'
Jan 12 16:51:56 chargestone-cave abrt-server[111301]: Can't open file 'core_backtrace' for reading: No such file or directory
Jan 12 16:51:57 chargestone-cave abrt-notification[111359]: Process 110460 (WebKitWebProcess) crashed in ??()
Jan 12 16:52:01 chargestone-cave abrtd[3593]: Size of '/var/spool/abrt' >= 5000 MB (MaxCrashReportsSize), deleting old directory 'ccpp-2019-01-12-16:51:55.522742-110460'
Jan 12 16:52:01 chargestone-cave abrt-server[111461]: Unsupported container technology
Jan 12 16:52:01 chargestone-cave abrt-server[111461]: Lock file '.lock' was locked by process 111464, but it crashed?
Jan 12 16:52:01 chargestone-cave abrt-server[111461]: Error: No segments found in coredump './coredump'
Jan 12 16:52:01 chargestone-cave abrt-server[111461]: Can't open file 'core_backtrace' for reading: No such file or directory
Jan 12 16:52:01 chargestone-cave abrt-notification[111519]: Process 111240 (WebKitWebProcess) crashed in ??()
[...]

PR #1481, opened 2020-05-04
Keep problem dirs locked while (core) backtrace is being generated

michalfabik:
See commits.  
  
Resolves https://github.com/abrt/abrt/issues/1357  
  
See also https://github.com/abrt/libreport/pull/629

### DIFF
--- a/src/daemon/abrtd.c
+++ b/src/daemon/abrtd.c
@@ -174,19 +174,13 @@ static void queue_post_create_process(struct abrt_server_proc *proc)
 
     char *worst_dir = NULL;
     const double max_size = 1024 * 1024 * abrt_g_settings_nMaxCrashReportsSize;
-    while (libreport_get_dirsize_find_largest_dir(abrt_g_settings_dump_location, &worst_dir, ignored) >= max_size
+    while (libreport_get_dirsize_find_largest_dir(abrt_g_settings_dump_location, &worst_dir, ignored, proc->dirname) >= max_size
            && worst_dir)
     {
         const char *kind = "old";
 
         GList *proc_of_deleted_item = NULL;
-        if (proc != NULL && strcmp(worst_dir, proc->dirname) == 0)
-        {
-            kind = "new";
-            stop_abrt_server(proc);
-            proc = NULL;
-        }
-        else if ((proc_of_deleted_item = g_list_find_custom(s_dir_queue, worst_dir, (GCompareFunc)abrt_server_compare_dirname)))
+        if ((proc_of_deleted_item = g_list_find_custom(s_dir_queue, worst_dir, (GCompareFunc)abrt_server_compare_dirname)))
         {
             kind = "unprocessed";
             struct abrt_server_proc *removed_proc = (struct abrt_server_proc *)proc_of_deleted_item->data;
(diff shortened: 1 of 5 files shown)


src/include/libabrt.h
```diff
@@ -44,7 +44,7 @@ void abrt_ensure_writable_dir_uid_gid(const char *dir, mode_t mode, uid_t uid, g
 void abrt_ensure_writable_dir(const char *dir, mode_t mode, const char *user);
 void abrt_ensure_writable_dir_group(const char *dir, mode_t mode, const char *user, const char *group);
 char *abrt_run_unstrip_n(const char *dump_dir_name, unsigned timeout_sec);
-char *abrt_get_backtrace(const char *dump_dir_name, unsigned timeout_sec, const char *debuginfo_dirs);
+char *abrt_get_backtrace(struct dump_dir *dd, const char *dump_dir_name, unsigned timeout_sec, const char *debuginfo_dirs);
```
mgrabovsky, line 47: Passing `dump_dir_name` is redundant now since it can be obtained from `dd`, isn'it?

src/plugins/abrt-action-generate-backtrace.c
```diff
@@ -78,7 +78,10 @@ int main(int argc, char **argv)
         debuginfo_dirs = libreport_xasprintf("%s:%s", debuginfo_location, i_opt);
 
     /* Create gdb backtrace */
-    char *backtrace = abrt_get_backtrace(dump_dir_name, exec_timeout_sec,
+    struct dump_dir *dd = dd_opendir(dump_dir_name, /*flags:*/ 0);
+    if (!dd)
+        return 1;
```
mgrabovsky, line 83: Would it make sense to log an error here?

src/daemon/abrtd.c
```diff
@@ -174,19 +174,13 @@ static void queue_post_create_process(struct abrt_server_proc *proc)
 
     char *worst_dir = NULL;
     const double max_size = 1024 * 1024 * abrt_g_settings_nMaxCrashReportsSize;
-    while (libreport_get_dirsize_find_largest_dir(abrt_g_settings_dump_location, &worst_dir, ignored) >= max_size
+    while (libreport_get_dirsize_find_largest_dir(abrt_g_settings_dump_location, &worst_dir, ignored, proc->dirname) >= max_size
```
mgrabovsky, line 177: Actually, what's the difference between the new argument and the one preceding it? Shouldn't that fulfill the same function? I may be missing something.

mgrabovsky:
The "Resolves" should probably point to #1357 instead.

src/plugins/abrt-action-generate-backtrace.c
michalfabik: Not really, IMO. `dd_opendir` and functions called therefrom already do that.

src/daemon/abrtd.c
michalfabik: That's what I thought too, but we want to ignore both the dir that was just created by the process that triggered the directory trimming and any directory from which a backtrace is being generated.

michalfabik:
> The "Resolves" should probably point to #1357 instead.

Ah, yes, those were duplicates and one was closed in the meantime.

src/plugins/abrt-action-generate-backtrace.c
mgrabovsky: Right. It's enough if `dd_opendir` complains in my opinion.

PR #1484, opened 2020-05-06
dbus: Update dd_get_owner

mgrabovsky:
Take over #1421.

Rebased onto master and changed return value of `abrt_p2_entry_get_owner` to 0 (`uid_t` may be unsigned).

Depends on abrt/libreport#630

### DIFF
--- a/src/dbus/abrt_problems2_entry.c
+++ b/src/dbus/abrt_problems2_entry.c
@@ -966,10 +966,16 @@ uid_t abrt_p2_entry_get_owner(AbrtP2Entry *entry,
     {
         g_set_error(error, G_DBUS_ERROR, G_DBUS_ERROR_IO_ERROR,
                     "Failed open dump directory");
-        return -1;
+        return (uid_t)-1;
     }
 
     const uid_t uid = dd_get_owner(dd);
+
+    if (errno != 0)
+        g_set_error(error, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                    "Failed to get owner of dump directory: %s",
+                    strerror(errno));
+
     dd_close(dd);
 
     return uid;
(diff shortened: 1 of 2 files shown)


src/dbus/abrt_problems2_entry.c
```diff
@@ -966,10 +966,16 @@ uid_t abrt_p2_entry_get_owner(AbrtP2Entry *entry,
...
+        return 0;
     }
 
-    const uid_t uid = dd_get_owner(dd);
+    int ret = 0;
+    const uid_t uid = dd_get_owner(dd, &ret);
+
+    if (ret != 0)
+        g_set_error(error, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                    "Failed to get owner of dump directory");
```
ernestask, line 977: Do you think consulting `strerror()` and adding the output to the message would be useful?

src/dbus/abrt_problems2_entry.c
```diff
@@ -966,10 +966,16 @@ uid_t abrt_p2_entry_get_owner(AbrtP2Entry *entry,
     {
         g_set_error(error, G_DBUS_ERROR, G_DBUS_ERROR_IO_ERROR,
                     "Failed open dump directory");
-        return -1;
+        return 0;
```
ernestask, line 969: This would be indistinguishable from root onwership, no? Maybe `abrt_p2_service_notify_entry_object()` should handle failures a bit better?

ernestask: Oh, wait, it’s below in the diff.

ernestask: Still, I think -1 would stand out better.

ernestask:
We should also bump the libreport dependency later.

PR #1485, opened 2020-05-06
setgid instead of setuid the abrt-action-install-debuginfo-to-abrt-cache [RHBZ 1796245]

xsuchy:


ernestask:
Have you tested this?

xsuchy:
Yes. Thou, additional testing is welcomed.

ernestask:
With the exception handler dropped:

```
Traceback (most recent call last):
  File "/usr/bin/abrt-action-install-debuginfo", line 323, in <module>
    sys.exit(run(config))
  File "/usr/bin/abrt-action-install-debuginfo", line 180, in run
    result = downloader.download(missing, download_exact_files=config.exact_fls)
  File "/usr/lib64/python3.8/site-packages/reportclient/debuginfo.py", line 336, in download
    retval = self.setup_tmp_dirs()
  File "/usr/lib64/python3.8/site-packages/reportclient/debuginfo.py", line 66, in wrapped
    os.setegid(abrt.pw_gid)
PermissionError: [Errno 1] Operation not permitted
```

xsuchy:
You need to have https://github.com/abrt/libreport/pull/631 as well.

ernestask:
```
$ grep ensure_abrt_gid /usr/lib64/python3.8/site-packages/reportclient/debuginfo.py
def ensure_abrt_gid(fn):
@ensure_abrt_gid
    @ensure_abrt_gid
```

It’s positively there. The call to `setegid()` never disappeared.

ernestask:
Do note that I’m running the binary as a regular user, not as abrt by proxy.

ernestask:
Ah, okay, that’s the issue. It never worked through `abrt-action-install-debuginfo` for regular users.

xsuchy:
I executed it using `abrt debuginfo-install ID`

ernestask:
Right, so this works fine for the usual case, but the concern I had was multi-user setups. The owner of the decompressed files will be the current user, with directories having 755 permissions. Any other user that would attempt to write there would be greeted with an error.

ernestask:
I can only imagine either having per-user caches or fixing this mess by turning all these binaries into D-Bus method calls, so that abrtd/abrt-dbus can properly manage everything.

src/plugins/abrt-action-install-debuginfo-to-abrt-cache.c
```diff
@@ -75,10 +75,9 @@ int main(int argc, char **argv)
     };
     const unsigned opts = libreport_parse_opts(argc, argv, program_options, program_usage_string);
 
+    const uid_t euid = geteuid();
```
ernestask, line 78: Is there a reason to move this line?

xsuchy: Nope. I moved it, because I removed it, but later I found that it is actually needed.

PR #1490, opened 2020-05-17
abrt-vmcore: Fix SyntaxWarning with python3.8

michalfabik:
python3.8 raises SyntaxWarning when comparing literals using `is` and it appears in `systemctl status abrt-vmcore`


ernestask:
Strictly speaking, there is a different reason why using `is` with literals is wrong, but this is fine.

Issue #1366, opened 2019-02-23
abrt.conf: tmpfiles should include more directories that should be created at runtime


pacho2:
abrt installs into paths that should be created at runtime. Probably current abrt.conf tmpfiles file should be extended to include the following dirs:
 *  / var/cache
 *   /var/cache/abrt-di
 *   /var/run/abrt

Thanks

ernestask:
> /var/cache

Uhh… No.

PR #1499, opened 2020-06-08
init-scripts: Add directories to systemd-tmpfiles config

ernestask:
This commit adds /var/cache/abrt-di and /var/run/abrt to the
systemd-tmpfiles config for automatic creation and, should the user want
it, timed cleanup.

Closes https://github.com/abrt/abrt/issues/1366

### DIFF
--- a/init-scripts/abrt.conf
+++ b/init-scripts/abrt.conf
@@ -1,3 +1,9 @@
-#Type Path          Mode UID  GID
-d     /var/tmp/abrt 0755 abrt abrt
+#Type Path                  Mode UID  GID
+d     /var/tmp/abrt         0755 abrt abrt
 x     /var/tmp/abrt/*
+
+d     /run/abrt             0755 root root
+r!    /run/abrt/abrt.pid
+r!    /run/abrt/abrt.socket
+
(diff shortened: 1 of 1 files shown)


ernestask:
Wasn’t sure about the spec file, so I just left the creation there as well, because my belief is that systemd-tmpfiles is only run periodically, not as part of some post-install scriptlet as well.

init-scripts/abrt.conf
```diff
@@ -1,3 +1,9 @@
-#Type Path          Mode UID  GID
-d     /var/tmp/abrt 0755 abrt abrt
+#Type Path                  Mode UID  GID
+d     /var/tmp/abrt         0755 abrt abrt
 x     /var/tmp/abrt/*
+
+d     /run/abrt             0755 root root
+x     /run/abrt/abrt.pid
+x     /run/abrt/abrt.socket
```
xsuchy, line 7: I would rather use 

    r!     /run/abrt/abrt.pid
    r!     /run/abrt/abrt.socket

ernestask: Yeah, that makes more sense.

ernestask: Changed.

PR #1502, opened 2020-06-22
Fix defects found by Coverity

mgrabovsky:
First batch of fixes.


src/dbus/abrt_problems2_service.c
```diff
@@ -2594,7 +2594,7 @@ int abrt_p2_service_register_objects(AbrtP2Service *service, GDBusConnection *co
 
     service->pv->p2srv_p2_object = abrt_p2_object_new(service,
                                                       &(service->pv->p2srv_p2_type),
-                                                      (char *)ABRT_P2_PATH,
+                                                      g_strdup(ABRT_P2_PATH),
```
ernestask, line 2597: It should really be “callee copies whatever is needed”, but the original authors had a different opinion. Oh well.

ernestask:
Seems fine to me.

PR #1503, opened 2020-06-22
Add GitHub Actions workflow for building PRs

ernestask:
So that we can get rid of the Jenkins job that is only accessible
within the Red Hat intranet.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,34 @@
+name: Build
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


ernestask:
We can do something fancier later on, I don’t have any quick ideas right now.

ernestask:
I’ll go ahead and do this for libreport and satyr, so I can finally just shut down that Jenkins instance.

ernestask:
Don’t mind if I don’t.

PR #1513, opened 2020-07-30
plugins: abrt-dump-journal-core: Handle zstd compression

ernestask:
Since v246-rc1, systemd supports zstd-compressed objects, which
definitely includes core dumps. This commit adds another check,
alongside .lz4 and .xz extensions in COREDUMP_FILENAME.

### DIFF
--- a/src/plugins/abrt-dump-journal-core.c
+++ b/src/plugins/abrt-dump-journal-core.c
@@ -264,21 +264,14 @@ save_systemd_coredump_in_dump_directory(struct dump_dir *dd, struct crash_info *
     if (coredump_path != abrt_journal_get_string_field(info->ci_journal, "COREDUMP_FILENAME", coredump_path))
         log_debug("Processing coredumpctl entry without a real file");
 
-    const size_t len = strlen(coredump_path);
-    if (   (len >= 3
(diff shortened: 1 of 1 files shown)


ernestask:
Needs https://github.com/abrt/libreport/pull/656 to land first.

mgrabovsky:
Testing this on Rawhide right now.

rpmbuild chokes on the trailing `%endif # with python3` in the spec file, but that's a separate problem.

ernestask:
> rpmbuild chokes on the trailing `%endif # with python3` in the spec file, but that's a separate problem.

Yeah, last I recall, one of the maintainers had a change of heart and that should no longer be an issue in some future RPM release.

mgrabovsky:
Yeah, seems to work fine. Is it mergeable already?

ernestask:
I suppose it is. There’s no hard requirement on the libreport changes, but this will start failing differently with older versions.

mgrabovsky:
So bump the dependency version? We're going to need it anyway because of #1484.

ernestask:
And make it uninstallable? How about we release libreport first?

mgrabovsky:
That's what I had in mind. Sorry I didn't make myself clear.

PR #1518, opened 2020-08-13
Minor code clean up

mgrabovsky:
* Decommission `libreport_list_free_with_free()` -- follow-up to abrt/libreport#660
* Purge commented-out code.

### DIFF
--- a/src/daemon/rpm.c
+++ b/src/daemon/rpm.c
@@ -75,15 +75,6 @@ void rpm_destroy()
     rpmFreeCrypto();
     rpmFreeMacros(NULL);
     rpmFreeRpmrc();
-
-/* rpm >= 4.14 handles this automatically on exit */
-#if 0
-    /* RPM doc says "clean up any open iterators and databases".
-     * Observed to eliminate these Berkeley DB warnings:
-     * "BDB2053 Freeing read locks for locker 0x1e0: 28718/139661746636736"
-     */
-    rpmdbCheckTerminate(1);
-#endif
 #endif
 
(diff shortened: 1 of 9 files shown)


src/daemon/rpm.c
```diff
@@ -75,19 +75,9 @@ void rpm_destroy()
...
-     * Observed to eliminate these Berkeley DB warnings:
-     * "BDB2053 Freeing read locks for locker 0x1e0: 28718/139661746636736"
-     */
-    rpmdbCheckTerminate(1);
-#endif
 #endif
 
-    libreport_list_free_with_free(list_fingerprints);
-    list_fingerprints = NULL;
+    g_list_free_full(g_steal_pointer(&list_fingerprints), free);
```
ernestask, line 80: Would `g_clear_list()` not be better in this case?

ernestask: Hmm, not for Fedora 31, at least.

ernestask:
Looks good overall, just that one question of dropping support for F31 prematurely.

src/daemon/rpm.c
mgrabovsky: There's always this one little function that I hadn't heard about which does that one more step.

Well, we could wait two months and “improve” the code even further once F31 is gone. I'm okay wit that for now.

ernestask: Not to brag (much), but it’s a rather new thing: https://gitlab.gnome.org/GNOME/glib/commit/58ba7d78fbd7ecb4c0df2dc7e251627ebbffb9d5

We can wait, I have no preference.

PR #1520, opened 2020-08-17
spec: Fix Obsoletes

michalfabik:
Koji builds fail because:
It's not recommended to have unversioned Obsoletes
It's not recommended to use '>' in Obsoletes

### DIFF
--- a/abrt.spec.in
+++ b/abrt.spec.in
@@ -265,7 +265,8 @@ Summary: %{name}'s pstore oops addon
 Requires: %{name} = %{version}-%{release}
 Requires: abrt-libs = %{version}-%{release}
 Requires: abrt-addon-kerneloops
-Obsoletes: abrt-addon-uefioops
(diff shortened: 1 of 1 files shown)


xsuchy:
You should also add there

    Provides: libreport-plugin-bodhi = %{version}

to handle clean upgrades.

See https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages

michalfabik:
> You should also add there
> 
> ```
> Provides: libreport-plugin-bodhi = %{version}
> ```

You mean:  
`Provides: abrt-addon-uefioops = %{version}-%{release}`?  
There already is a line like that for `libreport-plugin-bodhi` (see changes). Or do you actually mean without the `-%{release}`?

xsuchy:
Yeah. Bodhi has it. Just the uefioops is needed. The version is enough, but version-release makes no harm.

Issue #1528, opened 2020-09-23
abrt-action-generate-backtrace crashes during local processing 


HaasJona:
See https://bugzilla.redhat.com/show_bug.cgi?id=1881745
Abrt crash while processing a test crash (kill -SIGSEGV gedit)

```
Sep 22 19:03:52 fmac.local systemd-coredump[2860]: [🡕] Process 2847 (abrt-action-gen) of user 1000 dumped core.
Stack trace of thread 2847:
#0  0x00007ff37db6abc5 raise (libc.so.6 + 0x3dbc5)
#1  0x00007ff37db538a4 abort (libc.so.6 + 0x268a4)
#2  0x00007ff37dbad127 __libc_message (libc.so.6 + 0x80127)
#3  0x00007ff37dbb4e1c malloc_printerr (libc.so.6 + 0x87e1c)
#4  0x00007ff37dbb6674 _int_free (libc.so.6 + 0x89674)
#5  0x00007ff37dd6ab99 g_free (libglib-2.0.so.0 + 0x56b99)
#6  0x0000556a4dfb6539 main (abrt-action-generate-backtrace + 0x1539)
#7  0x00007ff37db551a2 __libc_start_main (libc.so.6 + 0x281a2)
#8  0x0000556a4dfb669e _start (abrt-action-generate-backtrace + 0x169e)
```

Suspected cause:
https://github.com/abrt/abrt/commit/89e0ddda37d8eb9b3aff78858be87255b36379dc#diff-a2b3bbd04cbb577c318d002d11a69171R351

This will (as I understand it) always free `bt` instead of only on error.

michalfabik:
Well, this truly is embarrassing. Thank you.

PR #1529, opened 2020-09-23
hooklib: Don't g_autofree backtrace

michalfabik:
The result of abrt_get_backtrace was being freed every time, even when
no error occured.

Resolves:
https://github.com/abrt/abrt/issues/1528
https://bugzilla.redhat.com/show_bug.cgi?id=1881745


mgrabovsky (approved):
Beat me to it! Looks good.

PR #1537, opened 2020-11-10
abrtd: Don't jump over g_autofree

michalfabik:
A `goto` jump over a pointer declaration with `g_autofree` causes an
invalid free at end of scope.

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1895660

### DIFF
--- a/src/daemon/abrtd.c
+++ b/src/daemon/abrtd.c
@@ -172,7 +172,7 @@ static void queue_post_create_process(struct abrt_server_proc *proc)
         /* Move behind '/' */
         ++ignored;
 
-    g_autofree char *worst_dir = NULL;
+    char *worst_dir = NULL;
     const double max_size = 1024 * 1024 * abrt_g_settings_nMaxCrashReportsSize;
     while (libreport_get_dirsize_find_largest_dir(abrt_g_settings_dump_location, &worst_dir, ignored, proc->dirname) >= max_size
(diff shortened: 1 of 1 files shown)


src/daemon/abrtd.c
```diff
@@ -192,6 +192,7 @@ static void queue_post_create_process(struct abrt_server_proc *proc)
                 abrt_g_settings_dump_location, abrt_g_settings_nMaxCrashReportsSize,
                 kind, worst_dir);
 
+        free(worst_dir);
         g_autofree char *deleted = g_build_filename(abrt_g_settings_dump_location ? abrt_g_settings_dump_location : "", worst_dir, NULL);
         worst_dir = NULL;
```
mgrabovsky, line 197: This doesn't look right. `worst_dir` is used in the call to `g_build_filename()`. I suggest `g_clear_pointer()` afterwards to perform both the free and nullifying at the same time.

mgrabovsky (changes requested):

src/daemon/abrtd.c
michalfabik: Of course! I looked at a wrong line of the diff with the original change and overlooked the arg to `g_build_filename`. Thanks!

mgrabovsky (approved):
Looks good now.

PR #1539, opened 2020-11-13
tests: Fix leaks

michalfabik:
Just tweak valgrind setup a bit and fix a few minor memory issues in tests. More errors that were popping up in the ABRT testsuite originated in Satyr code and were fixed in https://github.com/abrt/satyr/pull/310

### DIFF
--- a/tests/koops-parser.at
+++ b/tests/koops-parser.at
@@ -11,19 +11,22 @@ int run_test(const struct test_struct *test)
 {
 	FILE *fp = xfopen_ro(test->filename);
 	fprintf(stderr, "%s\t", test->filename);
-	g_autofree char *line = NULL;
-	g_autofree char *version = NULL;
+	char *line = NULL;
+	char *version = NULL;
 	while ((line = libreport_xmalloc_fgetline(fp)) != NULL)
 	{
 		version = abrt_koops_extract_version(line);
+        free(line);
 		if (version && !strcmp(version, test->expected_results))
 			break;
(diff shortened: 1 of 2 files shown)


mzidek-gh:
Hi,

few comments.

in the patch "tests: Only fail on definite and indirect leaks", can you give some explanation why failing on possible leaks caused problems (maybe with an example of false positive error)?

In the patch "tests: Fix leaks" can you please keep the initializations to NULL for line and version. With line it is not a big issue, but with version we could pass a garbage value to free if the while cycle does 0 iterations (if libreport_xmalloc_fgetline(fp) returns NULL in the first test).

Other then that LGTM.

tests/koops-parser.at
```diff
@@ -11,19 +11,22 @@ int run_test(const struct test_struct *test)
...
 	FILE *fp = xfopen_ro(test->filename);
 	fprintf(stderr, "%s\t", test->filename);
-	g_autofree char *line = NULL;
-	g_autofree char *version = NULL;
+	char *line;
+	char *version;
 	while ((line = libreport_xmalloc_fgetline(fp)) != NULL)
 	{
 		version = abrt_koops_extract_version(line);
+        free(line);
```
mgrabovsky, line 19: Minor nitpick, the file uses tabs for indentation while your new code uses spaces.

mzidek-gh (approved):

mzidek-gh:
ACK.

PR #1542, opened 2020-11-25
Use coredumpctl to retrieve coredump

michalfabik:
This PR moves the coredump retrieval to a later stage of the problem dir creation, namely by integrating it into `abrt-action-generate-backtrace` and `abrt-action-generate-core-backtrace`. The coredump is retrieved by calling `coredumpctl` directly, rather than by using `sd_journal_get_data`.

### DIFF
--- a/src/include/libabrt.h
+++ b/src/include/libabrt.h
@@ -45,6 +45,7 @@ void abrt_ensure_writable_dir(const char *dir, mode_t mode, const char *user);
 void abrt_ensure_writable_dir_group(const char *dir, mode_t mode, const char *user, const char *group);
 char *abrt_run_unstrip_n(const char *dump_dir_name, unsigned timeout_sec);
 char *abrt_get_backtrace(struct dump_dir *dd, unsigned timeout_sec, const char *debuginfo_dirs);
+void abrt_save_coredump(struct dump_dir *dd, unsigned timeout_sec);
 
 bool abrt_dir_is_in_dump_location(const char *dir_name);
 
--- a/src/plugins/abrt-action-generate-backtrace.c
+++ b/src/plugins/abrt-action-generate-backtrace.c
@@ -80,6 +80,12 @@ int main(int argc, char **argv)
     struct dump_dir *dd = dd_opendir(dump_dir_name, /*flags:*/ 0);
     if (!dd)
         return 1;
+
+    if (!dd_exist(dd, FILENAME_COREDUMP))
+    {
+        abrt_save_coredump(dd, exec_timeout_sec);
+    }
+
     g_autofree char *backtrace = abrt_get_backtrace(dd, exec_timeout_sec,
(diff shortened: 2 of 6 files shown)


src/include/libabrt.h
```diff
@@ -45,6 +45,7 @@ void abrt_ensure_writable_dir(const char *dir, mode_t mode, const char *user);
 void abrt_ensure_writable_dir_group(const char *dir, mode_t mode, const char *user, const char *group);
 char *abrt_run_unstrip_n(const char *dump_dir_name, unsigned timeout_sec);
 char *abrt_get_backtrace(struct dump_dir *dd, unsigned timeout_sec, const char *debuginfo_dirs);
+int abrt_save_coredump(struct dump_dir *dd, unsigned timeout_sec);
```
mzidek-gh, line 48: The return value of this function is never checked, nor is it actually needed. Can you please change it to void?

mzidek-gh: Or alternatively, check the return value for errors when you call it. Up to you. The only thing that could fail there is memory allocation, but with g_* familly of functions it will crash the program anyway, so I think void is ok.

src/plugins/abrt-action-generate-core-backtrace.c
```diff
@@ -63,17 +63,20 @@ int main(int argc, char **argv)
     g_autofree char *error_message = NULL;
     bool success;
 
+    struct dump_dir *dd = dd_opendir(dump_dir_name, /*flags:*/ 0);
+    if (!dd)
```
mzidek-gh, line 67: Please log an error here.

src/plugins/abrt-action-generate-backtrace.c
```diff
@@ -80,6 +80,10 @@ int main(int argc, char **argv)
     struct dump_dir *dd = dd_opendir(dump_dir_name, /*flags:*/ 0);
     if (!dd)
         return 1;
+
+    if (!dd_exist(dd, FILENAME_COREDUMP))
+        abrt_save_coredump(dd, exec_timeout_sec);
```
mzidek-gh, line 85: I think it is safer to use { } even for single commands in if-statements. However we do not have coding guidelines for that, so I will leave it to you.

michalfabik:
@mzidek-gh, issues resolved but please let's wait with this after the release.

mzidek-gh:
> @mzidek-gh, issues resolved but please let's wait with this after the release.

Thanks. Sure :)

msrb:
Scratch-build with these changes applied: https://koji.fedoraproject.org/koji/taskinfo?taskID=70681686

msrb (approved):
LGTM ;)

msrb:
Let's wait with merging a bit :)

michalfabik:
We've waited long enough. Merging, fingers crossed.

Issue #603, opened 2013-02-08
start versionin the libraries


mozeq:
We still have soname 0.0.1 and already ran into a few problems when changing API/ABI and not changing the soname.

xsuchy:
Related to https://github.com/abrt/libreport/issues/129

xsuchy:
note that in ABRT package we have:
/usr/lib64/libabrt_dbus.so.0.0.1  /usr/lib64/libabrt_gui.so.0.0.1  /usr/lib64/libabrt.so.0.0.1

PR #1548, opened 2020-12-09
libs: Add version info script

mzidek-gh:
Add version info script for liabrt and libart_gui.

Fixes:
https://github.com/abrt/abrt/issues/603

Nore: We will still have soname 0.0.1 until next release (we will bump the version for every release).

### DIFF
--- a/src/configuration-gui/Makefile.am
+++ b/src/configuration-gui/Makefile.am
@@ -11,7 +11,8 @@ libabrt_gui_la_SOURCES = \
 	abrt-config-widget.c \
 	abrt-config-widget.h \
 	system-config-abrt.c \
-	system-config-abrt.h
+	system-config-abrt.h \
+	libabrt_gui.sym
 
 # G_DEFINE_TYPE(...) macros result in
 # typedef '_GStaticAssertCompileTimeAssertion_0' locally defined but not used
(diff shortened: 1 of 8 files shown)


src/configuration-gui/Makefile.am
```diff
@@ -63,4 +68,4 @@ uidir = $(pkgdatadir)/ui
 EXTRA_DIST = abrt_gui.pc.in
 
 pkgconfigdir = $(libdir)/pkgconfig
-pkgconfig_DATA = abrt_gui.pc
```
mgrabovsky, line 66: Why was this removed?

mgrabovsky (commented):
Looks good to me. Only need a clarification above.

mgrabovsky:
Also, please add a changelog entry for the change.

src/configuration-gui/Makefile.am
mzidek-gh: This was some kind of strange git rebase error. I checked my local original patch this change was not there, but after rebasing the patch on top of current master the line was removed. Not sure why git did that, but it was not intended. Thanks for the catch.

mgrabovsky (approved):
Looks good to me.

PR #1551, opened 2021-01-03
Fix Coverity workflow

ernestask:
Are you tired of getting failure emails every weekend? Don’t answer that - a rhetorical question.

Then this is for you!

With the changes, the workflow itself finishes and can be triggered manually, but there is still the problem of Coverity Scan not supporting GCC 10 (though I forget the details now, but I recall complaining to them about some header-parsing thing). That one might take a year or two before they release a fixed-up version, given their cadence so far.


mgrabovsky (approved):
Looks good. Thanks!

PR #1553, opened 2021-01-11
Make rpm fix

michalfabik:
Fix a couple of minor issues that were causing `make rpm` to fail.  
  
See commits.


mgrabovsky (approved):
Looks good.

PR #1555, opened 2021-01-13
Add dependency on python3-systemd for abrt-addon-vmcore

mgrabovsky:
`abrt-action-check-oops-for-hw-error` from the abrt-addon-vmcore package imports `systemd.journal` but the python3-systemd package is not declared as a dependency, which causes the script to crash if python3-systemd is not installed on the system.
    
This commit fixes this by adding the formal dependency in the spec file.
    
Resolves [rhbz#1914170](https://bugzilla.redhat.com/show_bug.cgi?id=1914170)


michalfabik (approved):

michalfabik:
LGTM

Issue #1414, opened 2019-09-04
RPM 4.15 complains about comments after %endif in the spec file


ernestask:
See https://github.com/rpm-software-management/rpm/pull/625

Example output from rpmbuild:
```
RPM build errors:
    extra tokens at the end of %endif directive in line 81:  %endif # with python3

    extra tokens at the end of %endif directive in line 99:  %endif # with python3

    extra tokens at the end of %endif directive in line 136:  %endif # with python3

    extra tokens at the end of %endif directive in line 209:  %endif # with python3

    extra tokens at the end of %endif directive in line 268:  %endif # with python3

    line 280: It's not recommended to have unversioned Obsoletes: Obsoletes: abrt-addon-uefioops
    line 289: It's not recommended to use '>' in Obsoletes: Obsoletes: libreport-plugin-bodhi > 0.0.1
    extra tokens at the end of %endif directive in line 316:  %endif # with python3

    extra tokens at the end of %endif directive in line 349:  %endif # with python3

    extra tokens at the end of %endif directive in line 365:  %endif # with python3

    extra tokens at the end of %endif directive in line 405:  %endif # with python3

    extra tokens at the end of %endif directive in line 480:  %endif # with python3

    extra tokens at the end of %endif directive in line 603:  %endif # with python3

    extra tokens at the end of %endif directive in line 681:  %endif # with atomic

    extra tokens at the end of %endif directive in line 990:  %endif # with python3

    extra tokens at the end of %endif directive in line 1009:  %endif # with python3

    extra tokens at the end of %endif directive in line 1049:  %endif # with python3

    extra tokens at the end of %endif directive in line 1068:  %endif # with python3
```

ernestask:
I would really prefer to not “fix” this on our side, because treating comments like anything other than comments is silly.

ernestask:
Created https://github.com/rpm-software-management/rpm/issues/829.

ernestask:
Hm, big oops. Looks like the error was actually triggered by failing tests.

PR #1556, opened 2021-01-22
Removed comments in abrt.spec.in after %endif

mickgl:
Those comments caused syntax warning while running ./autogen.sh sysdeps on up-to-date Fedora 33.

> warning: extra tokens at the end of %endif directive in line 82:  %endif # with python3
>
> warning: extra tokens at the end of %endif directive in line 100:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 133:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 198:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 259:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 308:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 331:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 347:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 384:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 455:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 575:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 646:  %endif # with atomic
> 
> warning: extra tokens at the end of %endif directive in line 930:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 946:  %endif # with python3
> 
> warning: extra tokens at the end of %endif directive in line 991:  %endif # with python3

Generally speaking it's not a big deal, but it can be a bit confusing for someone who build ABRT from sources.

### DIFF
--- a/abrt.spec.in
+++ b/abrt.spec.in
@@ -79,7 +79,7 @@ BuildRequires: python3-devel
 BuildRequires: python3-systemd
 BuildRequires: python3-argcomplete
 BuildRequires: python3-dbus
-%endif # with python3
+%endif
 
 Requires: libreport >= %{libreport_ver}
 Requires: satyr >= %{satyr_ver}
@@ -97,7 +97,7 @@ Requires(pre): %{shadow_utils}
 %if %{with python3}
 Requires: python3-augeas
 Requires: python3-dbus
-%endif # with python3
+%endif
 %ifarch aarch64 i686 x86_64
 Requires: dmidecode
 %endif
@@ -130,7 +130,7 @@ BuildRequires: python3-sphinx
 BuildRequires: python3-libreport
 #python3-abrt-doc
 BuildRequires: python3-devel
-%endif # with python3
+%endif
 
 %description
 %{name} is a tool to help users to detect defects in applications and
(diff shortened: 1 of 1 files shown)


michalfabik:
Hi,  
  
there already was some back-and-forth concerning these:  

https://github.com/abrt/abrt/issues/1414  
https://github.com/rpm-software-management/rpm/issues/829  
  
I tend to side with @ernestask's reasoning that commenting-out a remainder of a line should be allowed and whatever tool complains about it needs to be fixed. On the other hand, I don't really feel that strongly about it so if you absolutely need the comments to go, I say why not.

ernestask:
So nothing has changed…

This was more an ideological battle for me, since I hardly ever find these comments useful, but you can re-evaluate at any point. The ABRT developers can be counted on, at most, two hands, after all.

mickgl:
Thanks for linking out already existing threads.
I know that comments are important and should be there (even if they are useful only for further developers), but from the other side, I found it really annoying. 
If we really want those comments to stay, why not move them one line above %endif?

michalfabik:
> If we really want those comments to stay, why not move them one line above %endif?  
  
I suppose that only makes them even more of a hassle when editing the file, moving lines around etc. But anyway, I don't really need them either, so unless anyone objects soon, I'll just merge this.

mzidek-gh:
Ok for me to merge.

michalfabik (approved):

mzidek-gh (approved):

PR #1557, opened 2021-02-09
Fixed rpm_load_gpgkey.

Daniil1918:
Added support of master + subkeys gpg.

### DIFF
--- a/src/daemon/rpm.c
+++ b/src/daemon/rpm.c
@@ -24,6 +24,18 @@
 #include <rpm/rpmcli.h>
 #include <rpm/rpmdb.h>
 #include <rpm/rpmpgp.h>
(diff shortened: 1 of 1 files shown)


mgrabovsky:
Hi, thanks for the pull request. Could you please explain the reasoning behind this change a bit more? Why is it necessary and what exactly does it fix?

Daniil1918:
> Hi, thanks for the pull request. Could you please explain the reasoning behind this change a bit more? Why is it necessary and what exactly does it fix?

When the file has a master public key and turnkey. And you need to read the fingerprint of both of them.

msrb:
Thanks for the pull request @Daniil1918 :wink: LGTM.

PR #1560, opened 2021-03-18
INSTALL: Add a hint on fixing failing testsuite

michalfabik:
Might trip up a user who's trying to build on a machine with ABRT
already installed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,6 +43,8 @@ otherwise run:
 
     $ make install
 
+(Note: If the build is failing, see if you don't have an older version of python3-abrt-addon already installed and try uninstalling the package as it might interfere with the build testsuite.)
+
 On platforms with SELinux enabled, the 'make install' command must be followed
 by the 'restorecon /' command. This is not required when installing rpm
 packages because rpm's selinux plugin takes care of it.


mgrabovsky:
Where does the interference in the tests come from?

michalfabik:
Hell if I knew. `tests/pyhook.at` sets path (`sys.path.insert(0, "../../../src/hooks")`) and then imports `abrt_exception_handler3`. `abrt_exception_handler3` is provided by `python3-abrt-addon`. When `python3-abrt-addon` is not installed, the test succeeds. When it is installed, the test fails. However, when the package is installed, `abrt_exception_handler3` gets imported from `/usr/lib/python3.9/` and when it's _not_ installed, it gets imported from `../../../src/hooks`. But then, what is the point of the `sys.path.insert`, or rather, why does it appear to not be doing anything? Moreover, modifying `tests/pyhook.at` in any way (e.g. to add `print(abrt_exception_handler3.__file__)` to see what's being imported) makes the tests fail too.  
  
In any case, https://github.com/abrt/abrt/pull/1562 seems to fix this.

mgrabovsky:
So it this OK to close or is it still relevant?

michalfabik:
I suppose so, yes.

Issue #1558, opened 2021-02-10
Wrong version displayed by CLI


mgrabovsky:
The `abrt` CLI displays the wrong package version:

```
$ rpm -q abrt-tui
abrt-tui-2.14.5-1.fc33.x86_64
$ abrt -v
abrt 2.14.2.9.gee974
```

michalfabik:
It seems `src/cli/abrtcli/config.py` doesn't necessarily get updated with every build. I kind of made it work by using `@PACKAGE_VERSION@` instead of `@VERSION@` in `Makefile.am` and `config.py.in` in the same directory. AFAIU, that var is set by `AC_INIT` in `configure.ac`. I'm not yet sure about the actual underlying cause though so I better not touch this until I understand it properly.

michalfabik:
Can't reproduce now:  
  
    $ sudo dnf remove abrt
    ...
    $ cd ~/wrk/abrt
    $ rm -r build/*
    $ git pull upstream master
    ...
    $ ./autogen.sh --prefix=/usr \
                       --sysconfdir=/etc \
                       --localstatedir=/var \
                       --sharedstatedir=/var/lib && make
    ...
    $ make rpm
    ...
    $ sudo dnf install -y --skip-broken build/noarch/* build/x86_64/*
    ...
    $ rpm -q abrt-tui
    abrt-tui-2.14.5.14.g0debf3-1.fc33.x86_64
    $ abrt --version
    abrt 2.14.5.14.g0debf3

mgrabovsky:
I think what you get from this sequence might be different from what is in the tarballs. Look into `src/cli/abrtcli/config.py` inside the [latest tarball](https://github.com/abrt/abrt/releases/tag/2.14.5):

    VERSION = "abrt 2.14.2.9.gee974"

michalfabik:
> I think what you get from this sequence might be different from what is in the tarballs  
  
Yeah, I've been suspecting that too. I'm basically just narrowing down where/when this might be happening.

michalfabik:
I tried `dnf downgrade` (yes, now it occurred to me) and again, the indicated version is one release older than the package version.  
  
    $ rpm -q abrt-tui
    abrt-tui-2.14.4-6.fc33.x86_64
    $ abrt --version
    abrt 2.14.2.9.gee974
    $  
  
It seems this has been happening consistently since devil knows when.

mzidek-gh:
This should be solved by @michalfabik's PR, changinf the assignee.
[...]

PR #1562, opened 2021-03-24
Use tito for releasing

michalfabik:
Clean up the autotools mess a little. Also, add a GitHub Workflow for automatic release creation. All lifted from libreport.

Resolves:
https://github.com/abrt/abrt/issues/1558


mzidek-gh:
Works for me. ACK.

mzidek-gh (approved):

PR #1567, opened 2021-05-07
Fix mismatched allocation / free

stevegrubb:
There are a number of places where memory was allocated with a glib2
function and then freed with the glibc free function. This patch
corrects the mismatch so that memory allocated by glib2 is freed by
glib2.


stevegrubb:
I have created a new glib2 model for Coverity which finds problms that were undetected before. I have a followup patch that fixes resource leaks.

michalfabik (approved):

PR #1568, opened 2021-05-12
Fix a few memory leaks

stevegrubb:
get_script_name allocates memory to script_pkg which must be freed.
If we assign script_pkg to pkg_name, then when it goes to ret, it
will call free_pkg_nevra on the memory to free it.

g_build_filename allocates memory to path that must be freed.
get_settings allocates memory to settings that must be freed

g_strdup allocates memory to exe. If it is not added to crash_info,
the it must be freed.


michalfabik:
Looking good, thanks!  
  
@mgrabovsky, I tested this yesterday and I'm somewhat intrigued by the fact that this doesn't make the results of `make maintainer-check-valgrind` any better. It might take me a while to look into what exactly is happening in the testsuite.

PR #1573, opened 2021-07-13
Update regex for matching interpreters

mgrabovsky:
Update regex for matching interpreter executables to match Python 3.10 as well as Perl 5.31.11.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1979762


michalfabik (approved):

michalfabik:
LGTM

Issue #1571, opened 2021-06-29
Port integration tests to tmt


mgrabovsky:
Our BeakerLib integration tests look ripe for porting to [tmt](https://tmt.readthedocs.io/en/stable/overview.html).

mgrabovsky:
First steps have been done in #1575.

PR #1575, opened 2021-07-29
Make it possible to run a subset of integration tests via TMT

msrb:
Not all tests work at the moment.


msrb:
@mgrabovsky PTAL :)

mgrabovsky:
I'll take a deeper look later today or tomorrow, but looks really lovely at a glance.

BTW, related (or resolves?): #1571

tests/runtests/prepare.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
+set -x
+
+dnf install -y \
```
mgrabovsky, line 6: So this assumes that the tests will only be run for released versions only, right? What if we wanted to test nightlies, for instance?

mgrabovsky (approved):
Looks sweet. All tests pass for me.

I only have one question, which is really a question rather than a critique.

tests/runtests/prepare.sh
msrb: We would need a different TMT "plan", which in this case is `integration.fmf`.
That plan could prepare the environment differently.

I am pretty sure it would be possible to parametrize the current plan/script as well, I am just not that familiar with `tmt` yet :)

msrb:
> BTW, related (or resolves?): #1571

Good catch. I'd say it's related, but there is still more work on the integration tests. I haven't tested all of them in `tmt`, but some of them are failing and will need some tweaking.

tests/runtests/prepare.sh
mgrabovsky: Got it, thanks.

HACKING.md
```diff
@@ -51,6 +51,16 @@ the tests/helpers directory.
...
 
+TL;DR: you can easily run a subset of the integration tests in a VM via [tmt](https://tmt.readthedocs.io/en/stable/).
+```shell
+# run tests in a Rawhide VM
+$ tmt run -vv
+# run tests in a F34 VM
+$ tmt run -vv --all provision --how virtual --image fedora-34
+```
+Note: You will need `tmt` and `tmt-provision-virtual` packages installed on your laptop.
+Also make sure that you have [virtualization configured](https://docs.fedoraproject.org/en-US/ci/tmt/#_virtualization_tips) properly.
```
psss, line 62: Actually, since `tmt-1.4` this is no more necessary, we're using qemu user mode so under regular user no additional actions should be needed.

tests/runtests/prepare.sh
psss: If you have fresh rpm packages built locally installing them on the guest is easy:

    tmt run prepare --how install --directory path/to/rpms

This however overrides config from plan so you might want to add a commented section in the plan like we do [here](https://github.com/psss/tmt/blob/master/plans/main.fmf#L15) and uncomment when desired.

mgrabovsky: Sweet, thanks.

PR #1576, opened 2021-07-30
Prevent unnecessary duplication of test metadata

psss:
Test metadata which are common for multiple tests can be inherited
so that it's not necessary to repeat the same data many times. The
`integration` plan moved one level higher to prevent inheriting as
well. Using `main.fmf` removes the extra suffix from test names.


mgrabovsky:
Looks great! Thanks for the improvements.

Could you please explain what you mean by “Using `main.fmf` removes the extra suffix from test names”?

psss:
> Could you please explain what you mean by “Using `main.fmf` removes the extra suffix from test names”?

In `fmf` filename becomes part of the object name, or `fmf id`. The `main.fmf` works in a similar way as `index.html`. To illustrate:

```
> tree
.
├── one
│   └── main.fmf
└── two
    └── test.fmf

> tmt test ls
/one
/two/test
```

By using `main.fmf` you get rid of the `test` suffix in the test name. Is it more clear now? See also fmf [names](https://fmf.readthedocs.io/en/latest/concept.html#names) documentation.

mgrabovsky:
@psss Oh, I get it now. Thanks for the explanation.

PR #1577, opened 2021-08-17
Enable more integration tests in TMT

msrb:


msrb:
There is some problem with `blacklisted-script` test though. I need to investigate little bit more.

msrb:
And suddenly:

`total: 2 tests passed, 4 tests failed and 19 errors`

Meh...

msrb:
I will need to check as the results were pretty flaky last time (IIRC).

PR #1579, opened 2021-09-23
cli: Fix path and glob matching for abrt info etc.

michalfabik:
Some command line tools were only accepting globs as args, not paths.
This was causing errors for relative paths such as '.'

This commit adds matching of actual paths before globs are tried.

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=2005659


src/cli/abrtcli/filtering.py
```diff
@@ -51,8 +51,15 @@ def filter_paths(problems, paths):
 
     def predicate(problem):
         for path in paths:
```
mgrabovsky, line 53: Could you please add a comment or two explaining what the code does? It's a bit confusing without any context at the moment.

michalfabik: It just checks whether any of the `problems` matches any of the `paths` where the latter can be contain both globs and paths. The fact that it's a function just makes it possible to use it with `filter()` a few lines later. I was thinking of renaming it to something more descriptive than "predicate" but then I noticed there's several more such functions throughout the file. I suppose I should come up with something universal. Or just rewrite the code without these functions. What do you think?

PR #1581, opened 2021-10-05
Do not try to free session data twice

msrb:
We free session data in on_g_signal() function, which is also
invoked when client disappears. Therefore, we don't need to register the
same free function in g_bus_watch_name_on_connection().

glib2 2.69.2 changed (fixed?) how/when g_bus_watch_name_on_connection()
calls the provided free function and it uncovered this problem in abrt-dbus.

See rhbz#1997315 for more details.


msrb:
@michalfabik @mgrabovsky PTAL :wink:

michalfabik (approved):
Makes sense, LGTM.

mgrabovsky (approved):
Looks good to me as well.

PR #1597, opened 2022-02-28
Check return values

mgrabovsky:
Check return values of various library functions and emit errors or warnings on failure.

### DIFF
--- a/src/applet/abrt-applet-application.c
+++ b/src/applet/abrt-applet-application.c
@@ -302,7 +302,8 @@ new_dir_exists (GList **new_dirs)
     cachedir = g_get_user_cache_dir ();
     dirlist_name = g_build_filename (cachedir, "abrt", NULL);
 
-    g_mkdir_with_parents (dirlist_name, 0777);
+    if (g_mkdir_with_parents (dirlist_name, 0777) != 0)
+        perror_msg_and_die (_("Could not create directory ‘%s’"), dirlist_name);
 
     g_free (dirlist_name);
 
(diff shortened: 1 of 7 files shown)


src/applet/abrt-applet-application.c
```diff
@@ -302,7 +302,8 @@ new_dir_exists (GList **new_dirs)
     cachedir = g_get_user_cache_dir ();
     dirlist_name = g_build_filename (cachedir, "abrt", NULL);
 
-    g_mkdir_with_parents (dirlist_name, 0777);
+    if (g_mkdir_with_parents (dirlist_name, 0777) != 0)
```
michalfabik, line 305: Off-topic (because I see you're just being consistent with the rest of the file) but is there a reason to put a space between the function name and the opening parenthesis?

mgrabovsky: It's only a matter of style. GLib (and GNOME/GTK?) use this convention and I think Ernestas wrote this code under that influence. The inconsistency with the rest of abrt makes me a wee bit uneasy, but I've learned to cope with it.

src/plugins/abrt-retrace-client.c
```diff
@@ -119,8 +119,9 @@ static int create_archive(bool unlink_temp)
     int tempfd = mkstemps(filename, /*suffixlen:*/7);
     if (tempfd == -1)
         perror_msg_and_die(_("Can't create temporary file in "LARGE_DATA_TMP_DIR));
-    if (unlink_temp)
-        g_unlink(filename);
+    if (unlink_temp && g_unlink(filename) != 0) {
```
michalfabik, line 122: FWIW, this file seems to be using a newline before the opening brace after an `if`.

mgrabovsky: Good catch. It's a mine field out there :roll_eyes:

PR #1598, opened 2022-03-03
a-a-generate-backtrace: Fail on missing coredump

michalfabik:
Don't save an empty coredump when coredump is no longer available to
coredumpctl. This typically happens when running
abrt-action-generate[-core]-backtrace by hand after the coredump has
been rotated away from /var/lib/systemd/coredump or, as the case may be,
when SELinux prevents the ABRT daemon from retrieving the coredump using coredumpctl.


mgrabovsky:
@michalfabik @msrb What's the status here? Is this still something we want merged? I can do the review if needed, once the conflicts are fixed.

michalfabik:
@mgrabovsky, no, I don't think we need this any more. This PR contains changes to code that @msrb has reverted in the meantime. Closing.

PR #1599, opened 2022-03-09
Revert coredump retrieval optimization patches (rhbz#2052872)

msrb:
This is the easiest way for now. Otherwise we would need to make sure that the SELinux problems are fixed not just in Fedora, but in other channels as well.


msrb:
Let's try this.

Issue #1604, opened 2022-04-07
abrt-oops: Cannot read journal data on f36


msrb:
abrt-oops seems to be failing on f36:

```
systemctl status abrt-oops.service 
× abrt-oops.service - ABRT kernel log watcher
     Loaded: loaded (/usr/lib/systemd/system/abrt-oops.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Wed 2022-04-06 19:14:00 CEST; 17h ago
   Main PID: 8827 (code=exited, status=1/FAILURE)
        CPU: 195ms

Apr 06 19:13:58 fedora systemd[1]: Started abrt-oops.service - ABRT kernel log watcher.
Apr 06 19:13:58 fedora abrt-dump-journal-oops[8827]: abrt-dump-journal-oops: Not restoring journal watch's posi>
Apr 06 19:14:00 fedora abrt-dump-journal-oops[8827]: abrt-dump-journal-oops: Failed to read 'MESSAGE' field: No>
Apr 06 19:14:00 fedora abrt-dump-journal-oops[8827]: abrt-dump-journal-oops: Cannot read journal data.
Apr 06 19:14:00 fedora systemd[1]: abrt-oops.service: Main process exited, code=exited, status=1/FAILURE
Apr 06 19:14:00 fedora systemd[1]: abrt-oops.service: Failed with result 'exit-code'.
```

PR #1605, opened 2022-04-08
abrt-oops: Follow journal from the end if the position cannot be restored

msrb:
See: https://github.com/abrt/abrt/issues/1604

The service would not start if the cursor state file was missing or invalid.
With this patch, the service will simply start following journal from the end.


mgrabovsky (approved):
Looks good.

PR #1609, opened 2022-05-27
retrace-client: Port to libsoup3

mcrha:
This is related to the effort of moving to libsoup3, tracked here:
https://gitlab.gnome.org/GNOME/libsoup/-/issues/218

and a Fedora discussion started here:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/JAQJ5WJQ6U6IZ3BZAZ5AM3VMMQCNOA7G/

Please note that the change needs testing, I do not know how you do that without touching real servers. The code only builds without warnings.


src/plugins/abrt-retrace-client.c
```diff
@@ -457,46 +496,50 @@ static char *get_release_id(GHashTable *osinfo, const char *architecture)
...
+        soup_message_headers_foreach(soup_message_get_response_headers(message), print_header, stderr);
     }
 
     if (response_code != 302 && response_code != 404)
     {
         alert_server_error(cfg.uri);
-        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%s"),
-                            response_code, response->data);
+        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%.*s"),
+                          response_code, response ? (int) g_bytes_get_size(response) : 0, response ? (const gchar *)g_bytes_get_data(response, NULL) : "");
```
mgrabovsky, line 542: As above, wrap this line, please.

src/plugins/abrt-retrace-client.c
```diff
@@ -841,11 +902,11 @@ static int create(SoupSession  *session,
     else if (response_code != 201)
     {
         alert_server_error(cfg.uri);
-        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%s"),
-                          response_code, response->data);
+        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%.*s"),
+                          response_code, response ? (int) g_bytes_get_size(response) : 0, response ? (const gchar *)g_bytes_get_data(response, NULL) : "");
```
mgrabovsky, line 906: Wrap this line, please.

src/plugins/abrt-retrace-client.c
```diff
@@ -886,112 +948,122 @@ static int run_create(SoupSession *session,
...
+        soup_message_headers_foreach(soup_message_get_response_headers(message), print_header, stderr);
     }
 
     if (response_code != 200)
     {
         alert_server_error(cfg.uri);
-        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%s"),
-                          response_code, response->data);
+        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%.*s"),
+                          response_code, response ? (int) g_bytes_get_size(response) : 0, response ? (const gchar *)g_bytes_get_data(response, NULL) : "");
```
mgrabovsky, line 991: Wrap this line, please.

src/plugins/abrt-retrace-client.c
```diff
@@ -886,112 +948,122 @@ static int run_create(SoupSession *session,
...
```
mgrabovsky, line 1055: Wrap this line, too, please.

src/plugins/abrt-retrace-client.c
```diff
@@ -1048,22 +1124,23 @@ static void exploitable(SoupSession  *session,
...
-        *exploitable_text = g_strdup(response->data);
+        *exploitable_text = g_strndup(g_bytes_get_data(response, NULL), g_bytes_get_size(response));
     }
     else
     {
         alert_server_error(cfg.uri);
-        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%s"),
-                          response_code, response->data);
+        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%*.s"),
+                          response_code, response ? (int) g_bytes_get_size(response) : 0, response ? (const gchar *)g_bytes_get_data(response, NULL) : "");
```
mgrabovsky, line 1133: Wrap this line, please.

src/plugins/abrt-retrace-client.c
```diff
@@ -1073,51 +1150,57 @@ static void run_exploitable(SoupSession *session,
...
+        soup_message_headers_foreach(soup_message_get_response_headers(message), print_header, stderr);
     }
 
     if (response_code != 200)
     {
         alert_server_error(cfg.uri);
-        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%s"),
-                          response_code, response->data);
+        error_msg_and_die(_("Unexpected HTTP response from server: %d\n%*.s"),
+                          response_code, response ? (int) g_bytes_get_size(response) : 0, response ? (const gchar *)g_bytes_get_data(response, NULL) : "");
```
mgrabovsky, line 1190: Wrap this line, please.

mgrabovsky (changes requested):
Fantastic! Thanks for the PR. The code looks fine at a glance, I've only pointed out a few exceedingly long lines. I'll test it functionally later today.

mcrha:
Sure, I can wrap those long lines. I'll wait for the functional testing before doing the changes, in case you'll find any problem(s) with the patch (it's likely).

By the way, thinking of it, is the retrace-client loaded into any other process or it runs on its own? I made this change to make gnome-shell run, but maybe the problem I faced was not related to the port. I'm not sure. If the retrace-server (aka the code loading libsoup) runs in the abrt space only, then this can be used (if working) directly, otherwise a compile time switch to use libsoup2 or libsoup3 will be needed, to match the libsoup version of the process the retrace-client is loaded into. I did not think of this option before.

mcrha:
I went ahead and made the requested changes. I wrapped few more lines. I did not wrap all exceeding the 85, because that would affect readability, from my point of view (see the "X-Task..." headers, only the header name would be on the next line, which is ugly).

My approach for the above claims was to split the lines and to change the indentation level. If you think it's a bad idea I can do the "use variables" approach, though the current one is neater.

mgrabovsky:
I'm so sorry, this has sort of fallen through the cracks. I'll make sure this moves forward today.

retrace-client always runs standalone so there should be no need for handling the older version.

mgrabovsky (approved):
Looks fine to me.

PR #1612, opened 2022-06-16
Fix for rpm 4.18

msrb:
`pgpHexStr()` function has been deprecated and renamed to `rpmhex()`.
    
Since the replacement is only available in Rawhide (f37), let's
continue using the old name for now.
    
This commit tells gcc to ignore the deprecation warning.
    
See the commit:
https://github.com/rpm-software-management/rpm/commit/d44be2cbc1c3265d55f05b47daa69334a7a133e6
    

msrb:
This commit has been backported to Rawhide already, so I am going to merge it :stuck_out_tongue:

PR #1615, opened 2022-08-11
spec: Fix macro expansion

michalfabik:
Not sure what the %{desktopvendor} macro does but if undefined, it
doesn't expand to anything without the '?', meaning the condition is
never true.


mgrabovsky (approved):
Good change.

PR #1616, opened 2022-08-23
lib: Use time_t for timestamps

mgrabovsky:
Use the appropriate standard type for timestamps instead of just `int` or `unsigned` in order to prevent any possible headaches down the road.

Problem reported by Coverity.

### DIFF
--- a/src/lib/hooklib.c
+++ b/src/lib/hooklib.c
@@ -121,12 +121,12 @@ static char* exec_vp(char **args, int redirect_stderr, int exec_timeout_sec, int
      * Therefore we have a (largish) timeout, after which we kill the child.
      */
     libreport_ndelay_on(pipeout[0]);
-    int t = time(NULL); /* int is enough, no need to use time_t */
-    int endtime = t + exec_timeout_sec;
+    time_t starttime = time(NULL);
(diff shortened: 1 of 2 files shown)


src/lib/hooklib.c
```diff
@@ -186,12 +186,12 @@ char *abrt_run_unstrip_n(const char *dump_dir_name, unsigned timeout_sec)
      * Therefore we have a (largish) timeout, after which we kill the child.
      */
     libreport_ndelay_on(pipeout[0]);
-    int t = time(NULL); /* int is enough, no need to use time_t */
-    int endtime = t + timeout_sec;
+    time_t starttime = time(NULL);
+    time_t endtime = starttime + timeout_sec;
```
michalfabik, line 190: Well to be completely pedantic, you could declare `timeout_sec` as `time_t` too in case somebody wanted to complain about the implicit conversion. But meh.

michalfabik:
:+1:

src/lib/hooklib.c
mgrabovsky: Yeah, I was thinking about that, too, but the timeout should always be a reasonably small number and the implicit conversion should be fine.

mgrabovsky:
Thanks for the review.

PR #1617, opened 2022-08-25
cli: use python3-argcomplete global completion

fredericmoulins:
Advertise support for argcomplete global completion. [1]

Remove the abrt specific bash completion script.

The bash completion script from python3-argcomplete in
`/etc/bash_completion.d/python-argcomplete`
will do the completion for `abrt`.

This change avoids the bash startup calling `register-python-argcomplete abrt`,
which takes some time.

[1]: https://github.com/kislyuk/argcomplete#global-completion

### DIFF
--- a/abrt.spec
+++ b/abrt.spec
@@ -947,7 +947,6 @@ killall abrt-dbus >/dev/null 2>&1 || :
 
 %files tui
 %if %{with python3}
-%config(noreplace) %{_sysconfdir}/bash_completion.d/abrt.bash_completion
 %{_bindir}/abrt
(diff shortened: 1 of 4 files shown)


michalfabik:
Hi, looks good to me, thanks! @mgrabovsky , WDYT?

mgrabovsky (approved):
Looks good to me as well. Thank you for the patch!

fredericmoulins:
Nice 👍

For the record, this change took off some 60ms off every interactive bash startup on Fedora on my (10+ year old) machine. I hope it reclaims some for everyone.

PR #1621, opened 2022-10-12
abrt-journal: call sd_journal_get_fd() right after sd_journal_open()

msrb:
See: rhbz#2128662

Under certain circumstances, abrt-dump-journal can be running, but not receiving any event notifications from journal.

The culprit of the issue seems to be the delayed call to sd_journal_get_fd(), as discussed in various
issues and pull-requests in other projects.
See for example [1], [2], or [3].

[1]: https://github.com/systemd/systemd/issues/7998
[2]: https://github.com/ledbettj/systemd-journal/pull/78
[3]: https://github.com/rsyslog/rsyslog/issues/2436

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   build:
     name: Build
-    container: fedora:latest
+    container: quay.io/fedora/fedora:latest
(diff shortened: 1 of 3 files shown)


michalfabik:
LGTM, great job figuring this out!

PR #1622, opened 2022-10-19
a-dump-journal-core: First seek the journal tail and then set filters

msrb:
See: rhbz#2128662

a-dump-journal-core uses filters to jump between
systemd-coredump entries in journal.

However, we should first jump to the starting position (journal tail) and only then set filters.
I suspect that setting the filters early, before we are at the starting position, can lead to the strange
behavior that we are seeing in Fedora.

This patch makes the problem go away on my freshly-installed Fedora 37 VM.

### DIFF
--- a/src/plugins/abrt-dump-journal-core.c
+++ b/src/plugins/abrt-dump-journal-core.c
@@ -628,16 +628,21 @@ main(int argc, char *argv[])
             error_msg_and_die(_("Cannot open systemd-journal"));
     }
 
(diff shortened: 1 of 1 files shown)


msrb:
Scratch: https://koji.fedoraproject.org/koji/taskinfo?taskID=93209160

mgrabovsky (approved):
Looks good to me. Thanks for the quick fix.

PR #1626, opened 2022-11-29
use SPDX format for license

xsuchy:
https://fedoraproject.org/wiki/Changes/SPDX_Licenses_Phase_1


mgrabovsky:
Looks good. Thanks for the patches.

Issue #1547, opened 2020-12-07
Rephrase Description of services


michalfabik:
The `Description` field in `init-scripts/abrt-<whatever>.service` currently says what the service _does_ rather than what it _is_. This produces systemd log entries such as: "Stopped Creates ABRT problems from coredumpctl messages." If this is the only place where these descriptions are used, it's probably a non-issue. Otherwise it might be worth fixing.

rquarry:
I believe this issue can be closed due to PR #1627

mgrabovsky:
Looks like it. Thanks for the heads up.

PR #1627, opened 2022-12-01
Update service descriptions

rquarry:
This pull request is in reference to issue #1547. Fairly simple update, although I'm not sure whether usage of "ABRT" or "Automated Bug Reporting Tool" is preferred for the description. Most .service files I looked at did not have acronnyms, so that route was chosen. 

Impressed with the docs on the project! Looking forward to learning more and contributing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- Update service descriptions in init scripts
+
 ## [2.16.0]
 ### Fixed
 - abrt-dump-journal-core: Fix the way how the service reads the journal (rhbz#2128662)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
```
mgrabovsky, line 8: Remove this empty line, please.

mgrabovsky (changes requested):
Thank you for the contribution, looks good overall. Could you please keep the “ABRT” in the messages, though? It can be interpreted as an acronym, but it's also sort intended not to be, and we almost never mention the “full” name, so I would stick with the wordmark for now.

CHANGELOG.md
rquarry: Thank you for the feedback-I'm not hitched to any specific format. Just to be clear by keeping the "ABRT" you mean changing the description to:

```Description=ABRT coredumpctl message creator```

or were you referring to simply tacking on "ABRT" to what I originally submitted? I think the latter would make the strings a bit longer than they should be.

mgrabovsky: Yes, the original, please.

rquarry: Just pushed a new commit with requested changes

mgrabovsky (approved):
Great, thanks for your contribution!

PR #1637, opened 2023-05-18
Replace now gone pgpHexStr() with rpmhex()

dmnks:
In RPM 4.19, the pgpHexStr() function was renamed to rpmhex(), much like
the comment here in the code suggests.  The function was deprecated in
RPM 4.18 while keeping the original shim.  That's no longer the case in
4.19, though, where it was removed.

Now that Fedora 38 is the latest stable release and Fedora 39 is getting
RPM 4.19, it's probably time to finally make the move.

### DIFF
--- a/src/daemon/rpm.c
+++ b/src/daemon/rpm.c
@@ -94,14 +94,6 @@ void rpm_destroy()
 }
 
 
-// TODO: pgpHexStr() has been renamed to rpmhex() recently, and
-//       pgpHexStr() is now deprecated and it will be dropped
-//       in the future. Let's keep using the old name for now
-//       as the replacement is only available in Rawhide (f37).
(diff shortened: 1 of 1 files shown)


dmnks:
Note that the GH actions failure here is caused by some other issue as the same happens when the Fedora package is build against Rawhide:
https://koji.fedoraproject.org/koji/taskinfo?taskID=101275801

mgrabovsky (approved):
Thanks for the patch. Looks good.

It looks like the failure is unrelated to this change.

dmnks:
Thanks!

How do we pull this fix into Fedora Rawhide? Should we just submit a PR against https://src.fedoraproject.org/rpms/abrt?

mgrabovsky:
We'll push the change downstream on next release. I suppose we'll have to fix the other error first anyway.

PR #1643, opened 2023-07-05
abrtd: remove livesys.service reference from the systemd unit file

msrb:
ABRT no longer uses /var/tmp/abrt/ for storing problem data. Therefore we are not concerned about livesys re-mounting /var/tmp/ as tmpfs and thus hiding it from abrtd.

Fixes: rhbz#2171200

### DIFF
--- a/init-scripts/abrtd.service
+++ b/init-scripts/abrtd.service
@@ -1,9 +1,5 @@
 [Unit]
 Description=ABRT Daemon
-# livesys.service has been added because of live distributions mounting tmpfs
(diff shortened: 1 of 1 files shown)


mgrabovsky (approved):

jwadodson:
It was mentioned that /var/tmp/abrt/  would not be created in future.

It still gets created in FC40 (2.17.6-1.fc40) - on reboot/restart?

If it's not required, when will it not be created?

jwadodson:
/var/tmp/abrt is still created in FC44 & recreated...

PR #1650, opened 2023-11-07
lib: kernel: Update kernel taint flag descriptions

marcan:
Mostly taken from Documentation/admin-guide/tainted-kernels.rst, with some minor wording changes or keeping the original wording where it was better.

In particular this adds X,T,N and completely changes the description of S to be more general.

Fixes S taint description on Apple Silicon systems (which get it due to a technicality related to KVM that definitely has nothing to do with SMP).


michalfabik:
Hi, thanks for the PR!

Issue #10, opened 2013-02-04
SELinux is preventing mailx (sendmail_t) "read" to ./binary_data1 (var_t).


mozeq:
Opened '20090226T12:48:56' by koca as https://fedorahosted.org/abrt/ticket/10

hi,[[BR]]
while attempting to send the report following SELinux preventing appeared[[BR]]
'''Detailed Description'''[[BR]]
SELinux denied access requested by mailx. It is not expected that this access is required by mailx and this access may signal an intrusion attempt. It is also possible that the specific version or configuration of the application is causing it to require additional access. [[BR]]
'''Allowing Access'''[[BR]]
Sometimes labeling problems can cause SELinux denials. You could try to restore the default system file context for ./binary_data1, restorecon -v './binary_data1'

Issue #1656, opened 2024-02-17
coredumpctl: core backtrace is useless


msrb:
This happens on aarch64:

```shell
           PID: 3219 (will_segfault)
           UID: 1000 (michal)
           GID: 1000 (michal)
        Signal: 11 (SEGV)
     Timestamp: Sat 2024-02-17 21:53:16 CET (2min 21s ago)
  Command Line: will_segfault
    Executable: /usr/bin/will_segfault
 Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-5bdca802-2875-4464-9245-0266a414b601.scope
          Unit: user@1000.service
     User Unit: vte-spawn-5bdca802-2875-4464-9245-0266a414b601.scope
         Slice: user-1000.slice
     Owner UID: 1000 (michal)
       Boot ID: 30727bd445334fc89dabad3a98c2e4d6
    Machine ID: 2a8ad4eeb2a74e22832e27803c6b1e09
      Hostname: fedora
       Storage: /var/lib/systemd/coredump/core.will_segfault.1000.30727bd445334fc89dabad3a98c2e4d6.3219.1708203196000000.zst (present)
  Size on Disk: 16.5K
       Package: will-crash/0.13.5-5.fc40
      build-id: 5aa9119f2f1080ede76b0b17fb0103a9b1393990
       Message: Process 3219 (will_segfault) of user 1000 dumped core.
                
                Module libwillcrash.so from rpm will-crash-0.13.5-5.fc40.aarch64
                Module will_segfault from rpm will-crash-0.13.5-5.fc40.aarch64
                Stack trace of thread 3219:
                #0  0x0000aaaaca9d02dc crash.constprop.0 (will_segfault + 0x102dc)
                #1  0x0070aaaaca9d0370 n/a (n/a + 0x0)
                #2  0x0070aaaaca9d0370 n/a (n/a + 0x0)
                #3  0x004faaaaca9d03bc n/a (n/a + 0x0)
                #4  0x005daaaaca9d03fc n/a (n/a + 0x0)
                #5  0x007effff8df1019c n/a (n/a + 0x0)
                #6  0x0075aaaaca9d029c n/a (n/a + 0x0)
                #7  0x0007aaaaca9d00ec n/a (n/a + 0x0)
                #8  0x0072ffff8dd409dc n/a (n/a + 0x0)
[...]

PR #1657, opened 2024-02-18
ccpp: don't fail bug reporting on ureport failure

msrb:
Related: gh#1656

ureport works with the core backtrace retrieved from coredumpctl, but the backtrace is sometimes quite useless:

    Stack trace of thread 3219:
    #0  0x0000aaaaca9d02dc crash.constprop.0 (will_segfault + 0x102dc)
    #1  0x0070aaaaca9d0370 n/a (n/a + 0x0)
    ...
    #9  0x0013ffff8dd40abc n/a (n/a + 0x0)
    #10 0x0076aaaaca9d0170 n/a (n/a + 0x0)
    ELF object binary architecture: AARCH64

abrt (understadably) complains about such low value backtrace and refuses to continue. However, when we later run gdb on the coredump, we actually get a good backtrace out of it.

This commit makes ureport failures non-blocking and thus there is a chance that we will get a good quality backtrace from gdb later.

### DIFF
--- a/src/plugins/ccpp_event.conf
+++ b/src/plugins/ccpp_event.conf
@@ -98,11 +98,11 @@ EVENT=report_Bugzilla type=CCpp duphash!=
 
 # Send micro report
 EVENT=report_uReport type=CCpp
-        /usr/libexec/abrt-action-ureport
+        /usr/libexec/abrt-action-ureport || :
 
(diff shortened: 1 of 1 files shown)


msrb:
Related: https://github.com/abrt/abrt/issues/1656

PR #1668, opened 2024-11-19
Fix typo in dbus error message

nvisser:
Fixed typo in src/dbus/abrt-dbus.c


msrb:
Good catch -- thank you for the pull request ;)

Issue #1690, opened 2026-01-25
Abrt is still trying to collect runlevel, which is nonsensical with systemd


ferdnyc:
In #630 (2013) it was requested that runlevel be added to the data collected by abrt.

Today `abrt_event.conf` still attempts to collect runlevel by running the `runlevel` command, which is not present on systemd-managed distros like Fedora 43. This leads to a line in the collected data like the following:

> runlevel:  `/bin/sh: line 1: runlevel: command not found`

The runlevel collection should either be removed, or made contingent on the command actually existing on the system.

ComicalSoftwares:
Hey, saw your issue.. thank you for reporting.

I tried fixing it, check out #1691

PR #1691, opened 2026-06-08
Fixed runlevel command not being found on systemd

ComicalSoftwares:
# What was the issue
As described by user `ferdnyc` in issue #1690, ABRT cannot run the command `runlevel` due to the command becoming deprecated on modern systemd.
# The solution I came up with
All I did was update the command inside the file `src/daemon/abrt_even.conf` at line **83**.. as you can see from the blame of the file, I pretty much just corrected and updated the commands.

#### Hope you accept my changes, I really wish the best for this awesome tool.

### DIFF
--- a/src/daemon/abrt_event.conf
+++ b/src/daemon/abrt_event.conf
@@ -80,7 +80,8 @@ EVENT=post-create remote!=1
 
 # Record runlevel (if not yet done) and don't return non-0 if it fails:
 EVENT=post-create runlevel= remote!=1
-        systemctl get-default >runlevel 2>&1
+        systemctl is-system-running >runlevel 2>&1
+        systemctl get-default >>runlevel 2>&1
         exit 0
 
 # A dummy EVENT=post-create for uploaded problems.
--- a/src/daemon/abrt_event.conf
+++ b/src/daemon/abrt_event.conf
@@ -80,8 +80,8 @@ EVENT=post-create remote!=1
 
 # Record runlevel (if not yet done) and don't return non-0 if it fails:
 EVENT=post-create runlevel= remote!=1
-    systemctl get-default >runlevel 2>&1
-    exit 0
+        systemctl get-default >runlevel 2>&1
+        exit 0
 
 # A dummy EVENT=post-create for uploaded problems.
 # abrtd would delete uploaded problems without this EVENT.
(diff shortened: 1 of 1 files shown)


src/daemon/abrt_event.conf
```diff
@@ -80,8 +80,8 @@ EVENT=post-create remote!=1
 
 # Record runlevel (if not yet done) and don't return non-0 if it fails:
 EVENT=post-create runlevel= remote!=1
-        runlevel >runlevel 2>&1
-        exit 0
+    systemctl get-default >runlevel 2>&1
+    exit 0
```
ferdnyc: I know one thing for sure, you definitely shouldn't change the code indentation for no reason.

```suggestion
        systemctl get-default >runlevel 2>&1
        exit 0
```

ferdnyc (changes requested):
Thanks for looking at this, @ComicalSoftwares! I'm curious to hear the maintainers' take on whether they consider `systemctl get-default` an equivalent replacement for `runlevel`, and whether they think it's worth collecting that info at all. Since:

1. Unlike `runlevel`, which reports the **current** runlevel, knowing the name of the default target doesn't actually say anything about whether that target is currently active on the system. If a user were to run `systemctl emergency` to activate `emergency.target`, `systemctl get-default` would still return the same value (usually `graphical.target`).
2. Even if you know the name of the default target, that doesn't tell you anything about what that target **contains**, since Systemd targets are much more flexible and customizable than runlevels were.

Perhaps `systemctl is-system-running` would be a more useful replacement for `runlevel`? That will report, in a word, whether the current system runtime state is `initializing`, `starting`, `running`, `degraded`, `maintenance`, `stopping`, `offline`, or `unknown`.

(There's also the separate question of whether, if `runlevel` is replaced with a `systemctl` command, the data field should also be renamed from `runlevel` to something else in the GUI / abrt data.)

I don't have any definite answers to any of these questions, but I do have one style nitpick regarding your changes...

ComicalSoftwares:
Oh yeah, that is actually a really smart way of doing it, so basically I went with the safe side and changed the file to do both commands `systemctl get-default` & `systemctl is-system-running`.. so like that in the report we can see what is the **current state of the machine** and **what was the machine supposed to do** because hey, there are headless servers operating in the terminal unlike our personal computers which operating using a graphical interface.

For the indentation, yeah I did a bit of an oopsie _(forgive me, I'm new to this I just wanted to help_) I didn't really see that, nice catch !

And for the GUI, I will take a look and see what has to be changed because I used to believe this current repository is just the backend behind ABRT and that the GUI is in the gnome-abrt repository.

Thank you for taking a look and correcting me on some crucial points @ferdnyc !

